PR #9, opened 2026-03-13
feat(metric-structure-table): implement metric structure creation logic

dusanparipovic:


src/input_gems_business_view.py
aoustry, line 1: I think it is better to just keep it in views.py

src/input_gems_business_view.py
```diff
@@ -0,0 +1,71 @@
...
+from gems.study.parsing import InputSystem, parse_yaml_components  # type: ignore[import-untyped]
+
+from src.calendar import Calendar
+from src.catalog import Catalog
+from src.metrics import BusinessViewConfig
+from src.simulation_table import SimulationTable
+from src.taxonomy import Taxonomy
+
+
+class InputGemsBusinessView:
```
aoustry: I don't really see the interest of this class with respect to the class ViewBuilder -> I think you can just rename this class ViewBuilder and move it in views.py.

src/views.py
```diff
@@ -12,53 +12,31 @@
...
 from dataclasses import dataclass
 
 import polars as pl
-from gemspy.gems.input_system import InputSystem  # type: ignore[import-not-found]
+from gems.study.parsing import InputSystem  # type: ignore[import-untyped]
 
-from location import LocationPorts
-from metrics import BusinessViewConfiguration
-from simulation import SimulationTable
+from src.metrics import BusinessViewConfig
```
aoustry: - [x] Could you remove the word "Business" from the dataclass and methods names? (ex: BusinessView => View). Sorry for that !

src/base_model.py
```diff
@@ -10,13 +10,10 @@
...
+from pydantic import BaseModel, ConfigDict
 
-@dataclass
-class LocationPorts:
-    """
-    Location/ports: set of str or None
-    """
 
-    location_ports: frozenset[str] | None
+class GEMSViewBuilderBaseModel(BaseModel):
```
aoustry: Please rename in ViewBuilderBasedModel and move it in views.py

src/metrics.py
```diff
@@ -10,108 +10,65 @@
...
+    In memory representation of the business_view_config.yml file.
     """
 
-    id: str
-    scope: Scope
-    catalogs: list[Catalog]
-    aggregation: Aggregation
-    metrics: list[tuple[str, str]]
+    def __init__(self, config_file_path: Path) -> None:
+        parsed = self._load_business_view_file(config_file_path)
```
aoustry: load_view_config_file

src/metrics.py
```diff
@@ -10,108 +10,65 @@
...
+            item.taxonomy_category for item in parsed.scope if item.taxonomy_category
+        )
+        self.calendar_id: str = next(item.calendar for item in parsed.scope if item.calendar)
+        self.catalog_ids: list[str] = [c.id for c in parsed.catalog]
+        self.time_aggregation: TimeAggregation | None = parsed.aggregation[0].time if parsed.aggregation else None
+        self.metrics: set[tuple[str, str]] = {
+            (parts[0], parts[1]) for m in parsed.metrics if len(parts := m.id.split(".", 1)) == 2
+        }
+
+    def _load_business_view_file(self, business_view_file_path: Path) -> BusinessViewData:
```
aoustry: load_view_config_file

src/input_gems_business_view.py
```diff
@@ -0,0 +1,71 @@
...
+        self,
+        input_data_path: Path,
+    ) -> None:
+        self._check_input_data_structure(input_data_path)
+        self.simulation_table = SimulationTable(next(input_data_path.glob("simulation_table*")))
+        self.calendar = Calendar(next(input_data_path.glob("calendar*")))
+        self.system = self._load_system(input_data_path)
+        self.taxonomy = Taxonomy(input_data_path / "taxonomy.yml")
+        self.business_view_config = BusinessViewConfig(input_data_path / "business_view_config.yml")
+        self.catalogs: list[Catalog] = [
```
aoustry: I wonder wether calendar and catalogs are needed at this level, if they appear in business_view_config

aoustry (changes requested):

src/base_model.py
dusanparipovic: I would rather like to keep renamed class inside separated file to avoid circular import. 
I faced with that problem right now.

src/input_gems_business_view.py
dusanparipovic: All right removed !

src/model_library.py
```diff
@@ -0,0 +1,156 @@
...
+    id: str
+    expression: str
+
+
+class ObjectiveContributionDef(ViewBuilderBasedModel):
+    id: str
+    expression: str
+
+
+class ModelDefinition(ViewBuilderBasedModel):
```
aoustry: Why not using the GemsPy Model class ? (->extend it ?)

gems_views_builder/library.py
```diff
@@ -0,0 +1,156 @@
...
+class LibraryData(ViewBuilderBasedModel):
+    """Root structure of a library yml file."""
+
+    id: str
+    description: str = ""
+    port_types: list[PortTypeDef] = Field(default=[], alias="port-types")
+    models: list[ModelDefinition] = []
+
+
+class ModelLibrary:
```
aoustry, line 72: Why not using the GemsPy Library class ? (extend it ?)

src/simulation_table.py
```diff
@@ -0,0 +1,120 @@
...
+    }
+)
+
+# Filtered simulation table has all SIMULATION_TABLE columns plus granular_date from the calendar join
+FILTERED_SIMULATION_TABLE_COLUMNS: frozenset[str] = SIMULATION_TABLE_COLUMNS | {"granular_date"}
+
+
+class SimulationTable:
+    """
+    In memory representation of the SIMULATION_TABLE
```
aoustry: Is it really in memory? What if the SimulationTable is a 100Go Apache Parquet file ?

src/simulation_table.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+        filtered_table = filtered_lazy.collect(engine="streaming")
+        filtered_table.write_csv(output_path)
+        return FilteredSimulationTable(output_path)
+
+
+class FilteredSimulationTable:
+    """
+    In-memory representation of a filtered SIMULATION_TABLE.
```
aoustry: same question as above: what if the SimulationTable is larger than RAM?

src/views.py
```diff
@@ -10,55 +10,95 @@
...
+            catalog: Catalog = self.view_config._load_current_catalog(catalog_id)
+            # # 2.2 Iterate over all metrics for this catalog
+            for metric_id in metrics:
+                try:
+                    metric: Metric = catalog.get_metric_by_id(metric_id)
+                except ValueError:
+                    continue  # # We should decide do we want to break process fully or continue with the next metric
+
+                # # 2.3 Build metric structure table
+                MetricStructureBuilder(
```
aoustry: Why exporting a csv ? MetricStructureTable is just a throughput, we could just have it in RAM and drop it no ?

src/views.py
```diff
@@ -10,55 +10,95 @@
...
+        system_path = next(input_data_path.glob("system*"))
+        return InputSystem.from_file(system_path)
+
+    def execute(self) -> None:
+        # # 1. Filter simulation table
+        filtered_simulation_table = self.simulation_table.filter_simulation_table(  # noqa: F841
+            self.calendar, self.input_data_path / "simulation_table_filtered.csv"
+        )
+        print(filtered_simulation_table)  # to avoid copilot unnecessary warnings
+        # # 2. Create metric structure table
```
aoustry: This loop won't _only_ create MetricStructureTable, it will also use it...

src/system.py
```diff
@@ -0,0 +1,102 @@
...
+from gems.study.parsing import parse_yaml_components
+
+
+class InputSystem(GemsInputSystem):  # type: ignore[misc]
+    """
+    Extends gems.study.parsing.InputSystem with component grouping by model type.
+    # We should consider to update InputSystem directly in GemsPy to avoid this class.
+    """
+
+    def model_post_init(self, __context: object) -> None:
```
aoustry: could you explain what is it please ?

src/input_system.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Extends gems.study.parsing.InputSystem with component grouping by model type.
+    # We should consider to update InputSystem directly in GemsPy to avoid this class.
+    """
+
+    def model_post_init(self, __context: object) -> None:
+        super().model_post_init(__context)
+        self._components_by_model: dict[str, list[str]] = self._build_components_by_model()
+
+    def _build_components_by_model(self) -> dict[str, list[str]]:
```
aoustry: Is it really about "building" components ? You just want to reference them by model no?

gems_views_builder/system.py
```diff
@@ -0,0 +1,102 @@
...
+        """Return all component ids for the given model type (e.g. 'generator' -> ['generator_A1', ...])."""
+        return self._components_by_model[model_id]
+
+    def get_component_by_id(self, component_id: str) -> GemsInputComponent:
+        for component in self.components:
+            if component.id == component_id:
+                return component
+        raise ValueError(f"Component {component_id} not found")
+
+    def _get_peer_component(self, component_id: str, port_id: str) -> str | None:
```
aoustry, line 62: We can keep it like this for now, but please open a Github issue to keep in mind that this could potentially be optimized by setting up a better storage structures for connections (using dict/heap rather than list my help I guess).

src/metrics.py
```diff
@@ -10,108 +10,175 @@
...
+                term.taxonomy_category
+            )  # # O(1) access insted of O(n) from pseudo code
+            for component_id in components_in_taxonomy_category:
+                # # Here will be applied filter with respect to properties of the component
+                # # Breakdown properties will be applied here
+
+                # # locating function
+                metric_location = self.system.locating_function(component_id, term.location_ports)
+                # # append to the output csv on disk (polars write)
+                append_row_polars(
```
aoustry: is it really efficient to write on disk each time? It look like you're opening/closing the file each time. Moreover, as said in another comment, I wonder whether we really need disk storage for MetricStructureTable. To be discussed, I'm open to better understand pros and cons.

src/model_library.py
```diff
@@ -0,0 +1,156 @@
...
+
+from pathlib import Path
+
+import yaml
+from pydantic import Field
+
+from src.base_model import ViewBuilderBasedModel
+
+# ---------------------------------------------------------------------------
+# Nested structures (parameters, variables, ports, etc.)
```
aoustry: These classes already exist in GemsPy no?

gems_views_builder/library.py
aoustry, line 1: rename in library.py please!

gems_views_builder/system.py
aoustry, line 1: rename in system.py please

gems_views_builder/taxonomy.py
```diff
@@ -9,15 +9,47 @@
...
+    )  # for now keep like this because taxonomy.yml used for testing isn't completed
+    variables: list[TaxonomyItem] = []
+    parameters: list[TaxonomyItem] = []
+    ports: list[TaxonomyItem] = []
+    constraints: list[TaxonomyItem] = []
+    extra_outputs: list[TaxonomyItem] = Field(default=[], alias="extra-outputs")
+    properties: list[TaxonomyItem] = []
+
+
+class TaxonomyData(ViewBuilderBasedModel):
```
aoustry, line 39: Why not just "Taxonomy" ?

resources/test_files/test_1/catalogs/catalog_1.yml
aoustry, line 1: As regard the name of the test folders, we could just use `test_1`, `test_2`... etc to uniformise the style?

aoustry (changes requested):
Very nice PR (a bit large)!

I have a lot of comments for discussion!

src/simulation_table.py
aoustry: I clarify my question: can we implement the class so that all the data is not loaded in memory? The idea being that the class is just an "access point".

src/metrics.py
aoustry: If we keep storage on disk, ApacheParquet will probably be a better option, as you suggested @dusanparipovic

src/simulation_table.py
dusanparipovic: Nice question. 
Actually that's what we do.

- If you look carefully at the constructor of class `Simulation Table` && `Filtered Simulation Table` we use function `scan_csv` instead of `read_csv`
- Main difference is that scan_csv is method which actually do that you have in your mind.
- This makes just pointer(like pointer in C++) to the file
One more fault from my side but now it's updated, since `filtered_lazy.collect(engine="streaming")` isn't stable function actually they can broke the RAM, so I would rather use singe to write results directly in csv. 
- How streaming works from high perspective:

1. put chunks of data into memory(RAM)  
2. perform your operations 

- But since its isn't stable I want to be sure that our algorithm is stable I will rather use sink function. 
[Stack Overflow](https://stackoverflow.com/questions/71021201/can-i-process-a-dataframe-using-polars-without-constructing-the-entire-output-in) 

- What could be used in further stages of optimization of our library [set_streaming_chunk_size](https://docs.pola.rs/api/python/dev/reference/api/polars.Config.set_streaming_chunk_size.html#polars.Config.set_streaming_chunk_size) 
- I need to do more research how we could adapt this for our usage and what will be boundaries for memory. Probably we could recommend something for our users in future and make this as option for usage.

src/simulation_table.py
dusanparipovic: [Answer](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/9#discussion_r2959604449)

src/views.py
dusanparipovic: I think you’re right, I’m also curious how this will behave in practice.

We want to perform a `join` and `group by` between:

* a simulation table stored on disk, and
* a metric structure table that can be either in memory or on disk

From a computer science perspective, accessing data in RAM is significantly faster than reading from disk. Because of that, the following scenarios are worth evaluating:

1. Simulation table on disk + metric structure table on disk
2. Simulation table on disk + metric structure table in memory

In theory, the second approach should be faster due to reduced I/O and quicker access to the smaller table.

However, the actual behavior depends on how libraries like Polars, Spark, or DuckDB optimize execution (e.g., predicate pushdown, join strategies, streaming, batching). Because of that, the best approach is to benchmark both scenarios and compare execution time.

Based on results, we can decide which strategy is more efficient for our use case.
But from my point of view second will be faster.

src/views.py
dusanparipovic: I will change comment to be more accurate.

src/metrics.py
dusanparipovic: [Reference](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/9#discussion_r2959637609)
I'll proceed with option 2 for now since I think you're definitely right METRIC STRUCTURE TABLE will be tiny object which could fit in memory. 
Thank you for your curiosity.

gems_views_builder/taxonomy.py
dusanparipovic: It will lead to name collision.

src/model_library.py
dusanparipovic: Yes indeed, thank you for that! 
I will adapt code,sorry for mistake.

src/input_system.py
dusanparipovic: Make sense,we could change name to `_group_component_ids_by_model_id`

src/system.py
dusanparipovic: This is constructor of current class which will call constructor of parent class and add other additional fields.

src/calendar.py
```diff
@@ -26,11 +24,48 @@ class Calendar:
...
+            raise ValueError(f"Calendar file {calendar_file_path} is not a CSV file")
+        return pl.scan_csv(calendar_file_path, try_parse_dates=True)
 
-    def abs_time_index_to_date(self, abs_time_index: int) -> datetime:
-        raise NotImplementedError()
+    def _check_calendar_columns(self) -> None:
+        df = self.dataframe.collect()
+
+        if df.columns != ["absolute_time_index", "block", "granular_date"]:
+            raise ValueError(f"Calendar '{self.id}' has invalid columns")
```
tbittar: Maybe sepcify which column is not here / not expected ?

src/base_model.py
```diff
@@ -10,13 +10,10 @@
...
 
-@dataclass
-class LocationPorts:
-    """
-    Location/ports: set of str or None
-    """
 
-    location_ports: frozenset[str] | None
+class ViewBuilderBasedModel(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
```
tbittar: From the docs : populate_by_name usage is not recommended in v2.11+ and will be deprecated in v3.
Instead, you should use the [validate_by_name][pydantic.config.ConfigDict.validate_by_name] configuration setting.
When validate_by_name=True and validate_by_alias=True, this is strictly equivalent to the
previous behavior of populate_by_name=True.

src/catalog.py
```diff
@@ -0,0 +1,77 @@
...
+    AVG = "avg"
+
+
+class TimeOperator(Enum):
+    SUM = "sum"
+    AVG = "avg"
+
+
+class Term(ViewBuilderBasedModel):
+    taxonomy_category: str = Field(alias="taxonomy-category")
```
tbittar: I think you can use a `alias-generator` in your Pydantic baseModel, see for instance : https://github.com/AntaresSimulatorTeam/GemsPy/blob/20d556a50dc764499b5b2ce3b4228acdcd6903cd/src/gems/utils.py#L81-L89

src/catalog.py
```diff
@@ -0,0 +1,77 @@
...
+    metrics_definition: list[Metric] = Field(alias="metrics-definition")
+
+
+class Catalog:
+    def __init__(self, catalog_file_path: Path) -> None:
+        parsed = self._load_catalog_file(catalog_file_path)
+        self.id = parsed.id
+        self.taxonomy = parsed.taxonomy
+        self.location_taxonomy_category = parsed.location.taxonomy_category
+        self.metrics: dict[str, Metric] = {metric.id: metric for metric in parsed.metrics_definition}
```
tbittar: Why do you need these "shortcuts" ?

src/catalog.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _load_catalog_file(self, catalog_file_path: Path) -> CatalogData:
+        with open(catalog_file_path) as f:
+            raw = yaml.safe_load(f)
+        return CatalogData.model_validate(raw["catalog"])
+
+    def get_metric_by_id(self, metric_id: str) -> Metric:
```
tbittar:
```suggestion
    def get_metric(self, metric_id: str) -> Metric:
```

src/input_system.py
```diff
@@ -0,0 +1,102 @@
...
+        groups: defaultdict[str, list[str]] = defaultdict(list)
+        for component in self.components:
+            model_ref = getattr(component, "model", None)
+            if not model_ref or "." not in model_ref:
+                continue
+            model_id = model_ref.split(".", 1)[1]
+            groups[model_id].append(component.id)
```
tbittar: Why not keeping model_ref as key ? In your case, if in different libs there are models with the same name, they will end up with the same key

src/system.py
```diff
@@ -0,0 +1,102 @@
...
+from collections import defaultdict
+from pathlib import Path
+from typing import Any, cast
+
+from gems.study.parsing import InputComponent as GemsInputComponent  # type: ignore[import-untyped]
+from gems.study.parsing import InputSystem as GemsInputSystem
+from gems.study.parsing import parse_yaml_components
+
+
+class InputSystem(GemsInputSystem):  # type: ignore[misc]
```
tbittar: @sylvlecl I think we need to avoid as much as possible this dependency for a easier integration later on ?
For now we can keep like it otherwise you need to "copy" the whole parsing lib of GEMSPy... but we need a "common parsing lib" for both this code and GEMSPy

src/input_system.py
tbittar: I suggest `self.models_to_components` (dict that maps models to components of a given model)

gems_views_builder/system.py
tbittar: This is already done if you use GEMSPy `System` class rather than `InputSystem` no ?

src/input_system.py
```diff
@@ -0,0 +1,102 @@
...
+            p1 = cast(str | None, getattr(conn, "port1", None))
+            c2 = cast(str | None, getattr(conn, "component2", None))
+            p2 = cast(str | None, getattr(conn, "port2", None))
+            if (c1, p1) == (component_id, port_id):
+                return c2
+            if (c2, p2) == (component_id, port_id):
+                return c1
+        return None
+
+    def locating_function(
```
tbittar: @aoustry
```suggestion
    def get_location(
```

gems_views_builder/metrics.py
```diff
@@ -10,108 +10,175 @@
...
 
-class TimeOperator(Enum):
-    """
-    Operator for time.
-    """
 
-    SUM = "sum"
-    AVG = "avg"
+class Scope(ViewBuilderBasedModel):
+    taxonomy_category: str | None = Field(None, alias="taxonomy-category")
```
tbittar, line 32: @aoustry do you want a location attribute here ? I understand this would only be necessary for later usage (geographic filtering) ?

gems_views_builder/metrics.py
```diff
@@ -10,108 +10,175 @@
...
-    catalogs: list[Catalog]
-    aggregation: Aggregation
-    metrics: list[tuple[str, str]]
+    def __init__(self, config_file_path: Path) -> None:
+        parsed = self._load_view_file(config_file_path)
+        self.id = parsed.id
+        self.location_taxonomy_category: str = next(
+            item.taxonomy_category for item in parsed.scope if item.taxonomy_category
+        )
+        self.calendar_id: str = next(item.calendar for item in parsed.scope if item.calendar)
```
tbittar: Why not just `parsed.scope.calendar` ? And same for `taxonomy_category` ? 

@aoustry It seems you may have several scope but you only keep one calendar / taxonomy_category, I do not understand if we can really have several scopes / or if keeping only one calendar / taxonomy_category is too restrictive

src/metrics.py
```diff
@@ -10,108 +10,175 @@
...
+        self.metrics: list[tuple[str, str]] = self._extract_metric_pairs(parsed.metrics)
+        # Internal helper: grouped by catalog
+        self.metrics_by_catalog: dict[str, list[str]] = self._group_metrics_by_catalog(self.metrics)
+
+    def _extract_metric_pairs(self, metrics: list[MetricRef]) -> list[tuple[str, str]]:
+        pairs: list[tuple[str, str]] = []
+        for metric in metrics:
+            catalog_id, metric_id = metric.id.split(".", 1)
+            pairs.append((catalog_id, metric_id))
+        return pairs
```
tbittar: Why not a dict `catalog-to-metrics` similarly as what you have done for models to components ?

tbittar: In fact this means merging `_extract_metric_pairs` and `_group_metrics_by_catalog`, so the question is rather why do you need to keep `self.metrics` ?

src/metrics.py
```diff
@@ -10,108 +10,175 @@
...
+        return metrics_for_catalog
+
+    def _load_view_file(self, view_file_path: Path) -> ViewData:
+        with open(view_file_path) as f:
+            raw = yaml.safe_load(f)
+        return ViewData.model_validate(raw["view"])
+
+    def _load_current_catalog(self, catalog_id: str) -> Catalog:
+        """
+        This method prevents us from having all catalogs in memory at once
```
tbittar: I do not think catalogs will be heavy in memory

gems_views_builder/metrics.py
```diff
@@ -10,108 +10,175 @@
...
+        with open(view_file_path) as f:
+            raw = yaml.safe_load(f)
+        return ViewData.model_validate(raw["view"])
+
+    def _load_current_catalog(self, catalog_id: str) -> Catalog:
+        """
+        This method prevents us from having all catalogs in memory at once
+        Just load when that is needed
+        """
+        if not (self.input_data_path / "catalogs" / f"{catalog_id}.yml").exists():
```
tbittar, line 101: input_data_path does not seem defined

src/metrics.py
```diff
@@ -10,108 +10,175 @@
...
+        """
+        if not (self.input_data_path / "calendars" / f"{self.calendar_id}.csv").exists():
+            raise FileNotFoundError(
+                f"Calendar file {self.input_data_path / 'calendars' / f'{self.calendar_id}.csv'} not found"
+            )
+        return Calendar(self.input_data_path / "calendars" / f"{self.calendar_id}.csv")
+
+
+class MetricStructureBuilder:
+    """InputSystem from GemsPy. LIST_OF_COMPONENTS_IN_TAXONOMY_CATEGORY, LOCATING_FUNCTION, build_tables (spec 2.1)."""
```
tbittar: InputSystem is rather your own class ?

src/metrics.py
```diff
@@ -10,108 +10,175 @@
...
+        Just load when that is needed,(e.g. when we're filtering simulation table)
+        """
+        if not (self.input_data_path / "calendars" / f"{self.calendar_id}.csv").exists():
+            raise FileNotFoundError(
+                f"Calendar file {self.input_data_path / 'calendars' / f'{self.calendar_id}.csv'} not found"
+            )
+        return Calendar(self.input_data_path / "calendars" / f"{self.calendar_id}.csv")
+
+
+class MetricStructureBuilder:
```
tbittar: Should we put it in a separate `metrics-builder.py` file ?

tbittar: So that this file only contains classes related to input data parsing and no "business" processing of these data

gems_views_builder/library.py
tbittar: This advocates for a common GEMS parsing lib reusable for different projects

src/simulation_table.py
tbittar: Does this work for both CSV and Parquet input format ?

src/taxonomy.py
```diff
@@ -9,15 +9,47 @@
...
     """
 
-    taxonomy_categories: dict[str, list[str]]
+    def __init__(self, taxonomy_file_path: Path) -> None:
+        with open(taxonomy_file_path) as f:
+            raw = yaml.safe_load(f)
+        parsed = TaxonomyData.model_validate(raw["taxonomy"])
+        self.id = parsed.id
+        self.description = parsed.description
+        self.categories: list[TaxonomyCategory] = parsed.categories
```
tbittar: What is the need of this ?

src/views.py
```diff
@@ -10,55 +10,95 @@
 #
 # This file is part of the Antares project.
 
-"""MetricStructureBuilder, BusinessViewBuilder, BusinessView (spec views.py)."""
+"""MetricStructureBuilder, ViewBuilder, View (spec views.py)."""
```
tbittar:
```suggestion
"""ViewBuilder."""
```

resources/test_files/test_1/catalogs/catalog_1.yml
dusanparipovic: Done!

src/base_model.py
dusanparipovic: Done!

src/calendar.py
dusanparipovic: Done!

src/catalog.py
dusanparipovic: Done !

src/input_system.py
dusanparipovic: Done!

src/input_system.py
dusanparipovic: Yes you're right,good catch.

gems_views_builder/system.py
dusanparipovic: Done!

src/catalog.py
dusanparipovic: see [`views.py`](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/blob/metric-structure-table/src/views.py) line 86

src/metrics.py
dusanparipovic: You're right, it’s not heavy, but it helped me process metrics by catalog and simplified the logic in the execute function inside `views.py`.
We can remove it if you prefer, but I find it useful since it’s just a small, three-line helper.

gems_views_builder/metrics.py
dusanparipovic: Right my mistake, solved !

src/model_library.py
dusanparipovic: Done !

src/model_library.py
dusanparipovic: Right done! 
One remark,added taxonomy-category since we don't have it inside GemsPy

gems_views_builder/library.py
dusanparipovic: I agree with Thomas, we can definitely consider that.

There’s quite a bit of repetitive boilerplate code, and the class structure isn’t very well organized. That’s probably why I overlooked that we already have a library model in GemsPy.

gems_views_builder/library.py
dusanparipovic: Done !

src/metrics.py
dusanparipovic: Yes but with tiny modification of InputSystem from GemsPy.
Core of the object is from GemsPy.

src/metrics.py
dusanparipovic: Fine for me. 
@aoustry What do you think ?

src/simulation_table.py
dusanparipovic: > Does this work for both CSV and Parquet input format ?

Yes [Polars docs](https://docs.pola.rs/api/python/stable/reference/api/polars.scan_parquet.html), with sink it should work correctly.

src/taxonomy.py
dusanparipovic: For now Antoine said that we can assume we have just list of categories, in future it will be adapted for more complex logic.

src/metrics.py
aoustry: @dusanparipovic @tbittar OK !

resources/test_files/test_1/simulation_table_daily_one_year.csv
sylvlecl, line 1: The files are **very** large for unit tests: couldn't we have smaller, more focused files ?

src/catalog.py
```diff
@@ -0,0 +1,76 @@
...
+# This file is part of the Antares project.
+
+"""In memory representation of a catalog .yml file."""
+
+from enum import Enum
+from pathlib import Path
+
+import yaml
+
+from src.base_model import ViewBuilderBasedModel
```
sylvlecl: Probably for later:
in the end we should have a parent namespace package named either `gems` or `antares`.
Imports should not need to include "src" (something to be configured in pyproject.toml or just in your IDE ?)

pyproject.toml
```diff
@@ -59,4 +68,8 @@ plugins = ["pydantic.mypy"]
 disable_error_code = ["attr-defined"]
 
 [[tool.mypy.overrides]]
-module = ["pydantic.*", "pytest.*","polars.*"]
\ No newline at end of file
+module = ["pydantic.*", "pytest.*", "polars.*"]
```
sylvlecl: Is it necessary, since we don't override any setting ?

pyproject.toml
```diff
@@ -59,4 +68,8 @@ plugins = ["pydantic.mypy"]
 disable_error_code = ["attr-defined"]
```
sylvlecl: we must re-enable that at some point

tests/test_catalog.py
```diff
@@ -0,0 +1,69 @@
...
+
+CATALOG_PATH = [
+    TEST_FILES_ROOT / "test_1" / "catalogs" / "catalog_1.yml",
+]
+
+
+@pytest.mark.parametrize("catalog_path", CATALOG_PATH)
+def test_catalog_loads(catalog_path: Path) -> None:
+    catalog = Catalog(catalog_path)
+    assert isinstance(catalog.id, str)
```
sylvlecl, line 28: In final tests we should check the actual value, not only the type which is more or less guaranteed by pydantic.
Same for all loading tests

src/simulation_table.py
sylvlecl: Good thing that you aimed from the start at handling larger than memory data.
Indeed the docstrings are a little misleading, would advise to remove the "in memory" wording to avoid confusion, and instead be explicit that we want to never load all the data in memory.

gems_views_builder/library.py
sylvlecl: Yes, the `parsing` layer of the gems library would better not be used outside from parsing stages of the process. They are just the plain image of the yaml format, the business logic should not rely on it directly.

Also, let's avoid use of inheritance when possible (mostly always :) ), let's use composition instead, it will be more explicit

src/library.py
```diff
@@ -0,0 +1,97 @@
...
+class ModelDefinition(GemsInputModel):  # type: ignore[misc]
+    """GemsPy `InputModel` plus optional `taxonomy-category` for ViewsBuilder indexing."""
+
+    taxonomy_category: str | None = Field(default=None, alias="taxonomy-category")
+
+
+class LibraryData(GemsInputLibrary):  # type: ignore[misc]
+    """Library root: GemsPy shape with extended model type."""
+
+    models: List[ModelDefinition] = Field(default_factory=list)
```
sylvlecl: See here for example, it's very hard to follow:
there is already a `models` field in the parent class, why do we override it here ?
Plus we need to go into the parent class to have the full picture.

Let's not use inheritance for such cases

src/catalog.py
```diff
@@ -0,0 +1,76 @@
...
+
+class CatalogData(ViewBuilderBasedModel):
+    id: str
+    taxonomy: str
+    location: CatalogLocation
+    metrics_definition: list[Metric]
+
+
+class Catalog:
+    def __init__(self, catalog_file_path: Path) -> None:
```
sylvlecl: Seems it would be more clear to have a plain dataclass and a loading method outside from it

src/calendar.py
```diff
@@ -26,11 +31,69 @@ class Calendar:
     3rd col: granular date
     """
 
-    id: str
-    dataframe: pl.DataFrame
+    def __init__(self, calendar_file_path: Path) -> None:
```
sylvlecl: same here, in the end it will be more clear to have an "as plain as possible" dataclass, and a loading method outside from it

src/taxonomy.py
```diff
@@ -9,15 +9,47 @@
...
 class Taxonomy:
     """
-    In memory representation of the taxonomy.yml file
-    Key: category name
-    Value: list of category values whose will be used as localization element.(In future, it will be a tree of categories.)
+    In memory representation of the taxonomy.yml file.
     """
 
-    taxonomy_categories: dict[str, list[str]]
+    def __init__(self, taxonomy_file_path: Path) -> None:
```
sylvlecl: same stuff about dissociating dataclass and parsing

src/system.py
```diff
@@ -0,0 +1,101 @@
...
```
sylvlecl: Let's avoid such inheritance and use composition instead.

sylvlecl (commented):
Nice work already.
I didn't have time to review thoroughly, just providing a few inputs for now:
as discussed earlier, most comments may be handled later if we want to have a first complete workflow quickly.

src/views.py
```diff
@@ -10,55 +10,92 @@
...
+
+    def execute(self) -> None:
+        # # 1. Filter simulation table
+        filtered_simulation_table = self.simulation_table.filter_simulation_table(  # noqa: F841
+            self.view_config.load_calendar(), self.input_data_path / "simulation_table_filtered.csv"
+        )
+        print(filtered_simulation_table)  # to avoid copilot unnecessary warnings
+        # # 2. Create metric structure table and perform SQL operations on it
+        # # Metrics are grouped by catalog, in order to prevent multiple loading of the same catalog
+        for catalog_id, metrics in self.view_config.metrics.items():
```
TheoPascoli: `self.view_config.metrics` is a list, you can not call `.items` on it. You should replace this by `self.view_config.metrics_by_catalog.items()`

gems_views_builder/metrics.py
```diff
@@ -10,108 +10,171 @@
...
+        self.time_aggregation: TimeAggregation | None = parsed.aggregation[0].time if parsed.aggregation else None
+        # Public API: list of (catalog_id, metric_id) pairs
+        self.metrics: list[tuple[str, str]] = self._extract_metric_pairs(parsed.metrics)
+        # Internal helper: grouped by catalog
+        self.metrics_by_catalog: dict[str, list[str]] = self._group_metrics_by_catalog(self.metrics)
+
+    def _extract_metric_pairs(self, metrics: list[MetricRef]) -> list[tuple[str, str]]:
+        pairs: list[tuple[str, str]] = []
+        for metric in metrics:
+            catalog_id, metric_id = metric.id.split(".", 1)
```
TheoPascoli, line 88: Are you sure you always have a `.` in your id ? You can add some validation before calling this before it will raise a `ValueError` if the id format is not correct

src/metrics.py
```diff
@@ -10,108 +10,171 @@
...
-    id: str
-    scope: Scope
-    catalogs: list[Catalog]
-    aggregation: Aggregation
-    metrics: list[tuple[str, str]]
+    def __init__(self, config_file_path: Path) -> None:
+        parsed = self._load_view_file(config_file_path)
+        self.input_data_path = config_file_path.parent
+        self.id = parsed.id
+        self.location_taxonomy_category: str = next(
```
TheoPascoli: If taxonomy_category does not contain any element, it will raise an `IterationError` which is not really understandable for the user. You should catch this case to tell the user that it forgot to write a taxonomy_category` in his file.
Something like:

```suggestion
        self.location_taxonomy_category = next(
      (item.taxonomy_category for item in parsed.scope if item.taxonomy_category),
      None,  # default value if empty                                                                 
  )
  if location_taxonomy_category is None:                                                                             
      raise ValueError(                                                                                              
          f"view_config '{parsed.id}': no 'taxonomy-category' found in scope. "
          f"At least one scope entry must define a taxonomy-category."                                               
      )                                                                                                              
  self.location_taxonomy_category = location_taxonomy_category 
```

src/system.py
```diff
@@ -0,0 +1,101 @@
...
+        for component in self.components:
+            model_ref = getattr(component, "model", None)
+            if not model_ref or "." not in model_ref:
+                continue
+            groups[model_ref].append(component.id)
+        return groups
+
+    def get_components(self, model_id: str) -> list[str]:
+        """Return all component ids for the given model type (e.g. 'generator' -> ['generator_A1', ...])."""
+        return self._components_by_model[model_id]
```
TheoPascoli: `KeyError` if the model does not exist. Replace it by .get(model_id, [])

tests/unit_tests/test_filter_simulation_table.py
```diff
@@ -0,0 +1,121 @@
...
+        .filter(pl.col("block") == pl.col("block_right"))
+        .collect(engine="streaming")
+        .height
+    )
+    assert filtered.height == expected_count, (
+        "Filtered row count must equal simulation rows whose (absolute_time_index, block) is in the calendar"
+    )
+
+
+@pytest.mark.parametrize("tmp_path", [None], indirect=True)
```
TheoPascoli: `tmp_path` is a builtin fixture, you can not use it with indirect. It works by accident but you can remove the decorator completely.

src/views.py
```diff
@@ -10,55 +10,92 @@
...
+    def _load_system(self, input_data_path: Path) -> InputSystem:
+        system_path = next(input_data_path.glob("system*"))
+        return InputSystem.from_file(system_path)
+
+    def execute(self) -> None:
+        # # 1. Filter simulation table
+        filtered_simulation_table = self.simulation_table.filter_simulation_table(  # noqa: F841
+            self.view_config.load_calendar(), self.input_data_path / "simulation_table_filtered.csv"
+        )
+        print(filtered_simulation_table)  # to avoid copilot unnecessary warnings
```
TheoPascoli: If you want to avoid copilot warnings, just write `_ = filtered_simulation_table` to not pollute production logs

src/views.py
```diff
@@ -10,55 +10,92 @@
...
+
+                # # 2.3 Build metric structure table
+                metric_structure_table = MetricStructureBuilder(
+                    self.system,
+                    catalog,
+                    metric,
+                    self.taxonomy,
+                    self.model_library,
+                ).build()
+                print(metric_structure_table)  # to avoid copilot unnecessary warnings
```
TheoPascoli: Same here.

```suggestion
                _ = metric_structure_table
```

src/views.py
```diff
@@ -10,55 +10,92 @@
...
+        # # 1. Filter simulation table
+        filtered_simulation_table = self.simulation_table.filter_simulation_table(  # noqa: F841
+            self.view_config.load_calendar(), self.input_data_path / "simulation_table_filtered.csv"
+        )
+        print(filtered_simulation_table)  # to avoid copilot unnecessary warnings
+        # # 2. Create metric structure table and perform SQL operations on it
+        # # Metrics are grouped by catalog, in order to prevent multiple loading of the same catalog
+        for catalog_id, metrics in self.view_config.metrics.items():
+            # # 2.1 Load catalog
+            catalog: Catalog = self.view_config._load_current_catalog(catalog_id)
```
TheoPascoli: You are calling a private method from outside. Either the method is public or ViewBuilder should not call it directly

gems_views_builder/system.py
TheoPascoli: I agree with @aoustry, you definitly should use a dict that you build in `model_post_init` because here you iterate `components * connections` times. Same for `get_component_by_id`

pyproject.toml
```diff
@@ -7,7 +7,13 @@ name = "GEMS-ViewsBuilder"
 version = "0.0.1"
 readme = "README.md"
 license = {file = "LICENSE"}
-dependencies = ["polars>=0.20.0","pyyaml==6.0.2", "pydantic>=2.11.7","PyYAML>=6.0.2",]
+dependencies = [
+    "polars>=1.39.2",
+    "pyyaml==6.0.2",
+    "types-PyYAML>=6.0.0",
```
TheoPascoli: `types-PyYAML` is not a runtime dep, you can move it to requirements-dev

src/library.py
```diff
@@ -0,0 +1,97 @@
...
+        self.models_by_taxonomy_category: dict[str, list[str]] = {}
+        for m in parsed.models:
+            if not m.taxonomy_category:
+                continue
+            self.models_by_taxonomy_category.setdefault(m.taxonomy_category, []).append(m.id)
+
+    def _load_library_file(self, library_file_path: Path) -> LibraryData:
+        with open(library_file_path) as f:
+            raw = yaml.safe_load(f)
+        return cast(LibraryData, LibraryData.model_validate(raw["library"]))
```
TheoPascoli: No need to cast here

gems_views_builder/catalog.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class TermsOperator(Enum):
+    SUM = "sum"
+    AVG = "avg"
+
+
+class TimeOperator(Enum):
+    SUM = "sum"
+    AVG = "avg"
```
TheoPascoli, line 36: Those 2 enums are the same. Can you merge them ? If you cant, you should document why you have two of them

src/catalog.py
```diff
@@ -0,0 +1,76 @@
...
+
+class TimeOperator(Enum):
+    SUM = "sum"
+    AVG = "avg"
+
+
+class Term(ViewBuilderBasedModel):
+    taxonomy_category: str
+    output_id: str
+    location_ports: str | None
```
TheoPascoli: Here, `location_ports` is a `str | None`. But the method `get_location` can also handle tuple. Its not a bug at the moment but this incoherence can lead to bugs

src/calendar.py
```diff
@@ -26,11 +31,69 @@ class Calendar:
...
+                parts.append(f"missing columns: {missing}")
+            if unexpected:
+                parts.append(f"unexpected columns: {unexpected}")
+            if not missing and not unexpected:
+                parts.append(f"wrong column order: expected {expected_columns}, got {actual}")
+            else:
+                parts.append(f"actual columns: {actual}")
+            raise ValueError(f"Calendar '{self.id}' has invalid columns: {'; '.join(parts)}")
+
+        df = self.dataframe.collect()
```
TheoPascoli: `self.dataframe` is a `LazyFrame`, but `_check_calendar_columns()` calls `self.dataframe.collect()` during construction, which loads the entire file into memory and defeats lazy evaluation.

Additionally, `scan_csv` already reads the header to infer the schema. Calling `pl.read_csv(n_rows=0)` reopens the file unnecessarily, `elf.dataframe.collect_schema().names()` provides the column names without additional disk access.

src/catalog.py
dusanparipovic: Agree, it will be done in this PR.

TheoPascoli (changes requested):
Good work overall ! The lazy loading approach is the right call for large files.

  A few things to fix before merging:
  - as mentioned in a previous comment, prefer composition over inheritance (especially `InputSystem` and `ModelLibrary`)
  - the new `MetricStructureBuilder` has no tests at all, and existing tests only cover the happy path, error cases (invalid file, missing column, malformed metric id, etc.) are not tested
  - 129 MB of test data committed to git (`simulation_table_daily_one_year.csv`, `simulation_table_hourly_one_week.csv`). Git history is permanent, these files cannot be cleanly removed after merge and will slow down every clone and CI checkout indefinitely. Use small synthetic fixtures instead.

resources/test_files/test_1/simulation_table_daily_one_year.csv
dusanparipovic: Agree. 
I will keep test_3 and adapt unit tests. 
I will provide smaller tests which we use as `end to end` test inside PyPSA to GEMS converter. 
Probably we could consider to have dedicated repository where we will store large scale tests for this library. 
Similar as we have [ANTARES_NR](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR) 
@sylvlecl

gems_views_builder/catalog.py
dusanparipovic: It can be merged now, but I’m not sure the two enums will remain identical in the future. 
I’d keep this approach for now to avoid duplicated work if we need to separate them later.
In fact they represent different operators.
If you agree ofc.

src/catalog.py
dusanparipovic: Right, nice catch!
Done!

src/catalog.py
dusanparipovic: Done!

src/simulation_table.py
dusanparipovic: Completely agree,done!

gems_views_builder/metrics.py
dusanparipovic: Yes by POC it will be always written catalog_id.metric_id. 
You're completely right,done!

pyproject.toml
dusanparipovic: Done!

gems_views_builder/catalog.py
TheoPascoli: Alright !

aoustry (dismissed):

TheoPascoli (approved):

aoustry (approved):

Issue #13, opened 2026-03-18
open question: decide where to place catalog and view-config files inside study folder


aoustry:

dusanparipovic:
Current input structure for Gems View Builder:
├── catalogs/ -> one or more catalog here 
├── view_config.yml
├── taxonomy.yml
├── system.yml
├── simulation_table.parquet
├── library.yml
└── calendar_file.csv

PR #20, opened 2026-03-27
feat(sql): add sql operation to complete workflow 

dusanparipovic:
**Summary**

- Implements full workflow
- Switches CSV to Parquet (for performance and future benchmarking)
- Adds new tests
- Optimizes `get_location`
- Adds PyPSA -> GEMS - based test file (adapted for this repo)
- Applies configuration to test files (simulation table Parquet reflects these changes)
- Merged chunks of files in order to get single final file e.g. `view`

**Parquet Configuration**

- `ZSTD compression` (default in Polars, likely similar in Spark/DuckDB/PyArrow, to confirm)
- `Data page version`: 2.0
- `Row group size`: 64,000 (open to tuning)
- `Compression level`: 3 (open to adjustment)

**Notes**

- Polars does not support sorting column metadata
- Workaround: convert to PyArrow before writing
- Sorted by:
- `absolute_time_index`, `block_time_index`, `scenario_index`,` component, output`


gems_views_builder/system.py
```diff
@@ -49,34 +50,32 @@ def models_to_components(self) -> dict[str, list[str]]:
             groups[model_ref].append(component.id)
         return groups
 
-    def get_components(self, model_id: str) -> list[str]:
-        """Return all component ids for the given model type (e.g. 'generator' -> ['generator_A1', ...])."""
-        return self._components_by_model.get(model_id, [])
+    def build_component_port_connections(self) -> dict[tuple[str, str], set[str]]:
```
aoustry, line 58: Could you a short docstring for this method please ?

gems_views_builder/system.py
```diff
@@ -49,34 +50,32 @@ def models_to_components(self) -> dict[str, list[str]]:
...
+        component_port_connections: dict[tuple[str, str], set[str]] = defaultdict(set)
+        for connection in self.connections:
```
aoustry, line 63: I suggest adding the following checks:

- component1!=component2
- component1 and component2 are not None
- port1 is indeed a port of component1's model (and idem for port2)

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
         system_path = next(input_data_path.glob("system*"))
         return InputSystem.from_file(system_path)
 
-    def execute(self) -> None:
-        # # 1. Filter simulation table
-        filtered_simulation_table = self.simulation_table.filter_simulation_table(  # noqa: F841
-            self.view_config.load_calendar(), self.input_data_path / "simulation_table_filtered.csv"
+    def _build_metric_view(self, joined_dataframe: pl.LazyFrame, metric: Metric) -> Path:
+        value_agg = pl.col("value").sum() if metric.terms_operator == TermsOperator.SUM else pl.col("value").mean()
```
aoustry: in the future we could have other aggregations operators -> you have to dispatch sum/mean and raise an "NotImplementedError" if the value is not available. A possible approach could just be to use a mapping (dict ?) between the possible TermsOperator and the corresponding polar/spark/duckdb operators

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
...
+        )
+        out_dir = self.input_data_path / "views" / "metric_view"
+        out_dir.mkdir(parents=True, exist_ok=True)
+        out_path = out_dir / f"{metric.id}.parquet"
+        metric_view.sink_parquet(out_path)
+        return out_path
+
+    def _build_view(self, metric_view_parquet_path: Path, metric: Metric) -> Path:
+        metric_view = pl.scan_parquet(metric_view_parquet_path)
+        time_agg = (
```
aoustry, line 182: same comment as line 79

gems_views_builder/views.py
```diff
@@ -38,7 +40,8 @@ def __init__(
         self.system = self._load_system(input_data_path)
         self.taxonomy = load_taxonomy(input_data_path / "taxonomy.yml")
         self.view_config = ViewConfig(input_data_path / "view_config.yml")
-        simulation_table_path = next(input_data_path.glob("simulation_table*"), None)
+        simulation_table_candidates = sorted(input_data_path.glob("simulation_table*.parquet"))
+        simulation_table_path = next((p for p in simulation_table_candidates if "filtered" not in p.stem), None)
```
aoustry: What is the behaviour? It tackles the first files of the list?

gems_views_builder/views.py
aoustry: Is value_agg a dataframe, or just the defition of an aggregation operator?

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
...
+                ]
+            )
+        )
+        out_dir = self.input_data_path / "views" / "metric_view"
+        out_dir.mkdir(parents=True, exist_ok=True)
+        out_path = out_dir / f"{metric.id}.parquet"
+        metric_view.sink_parquet(out_path)
+        return out_path
+
+    def _build_view(self, metric_view_parquet_path: Path, metric: Metric) -> Path:
```
aoustry: I think that the name of the method is not very explicit: we should clarify that it performs a temporal aggregation (for a given metric)

gems_views_builder/views.py
```diff
@@ -90,15 +169,43 @@ def execute(self) -> None:
...
+                    metric_structure_path
+                )  # # TO DO for benchmark: test behavior when using write_parquet and not sink_parquet since metric structure table won't be heavy datum
+                # # Apply sylvan suggestions for creation of new parquet file(meta data)
+
+                filtered_lazy = pl.scan_parquet(filtered_simulation_table_path)
+                metric_structure_lazy = pl.scan_parquet(metric_structure_path)
+
+                joined_dataframe = filtered_lazy.join(
+                    metric_structure_lazy,
+                    left_on=["component", "output"],
```
aoustry: we should maybe unify the notation between metric_structure_table and simulation_table as regards the columns names `component` and `output`?

gems_views_builder/views.py
```diff
@@ -90,15 +169,43 @@ def execute(self) -> None:
...
+                    right_on=["component_id", "output_id"],
+                    how="right",
+                )
+
+                metric_view_parquet_path = self._build_metric_view(joined_dataframe=joined_dataframe, metric=metric)
+                view_parquet_path = self._build_view(metric_view_parquet_path=metric_view_parquet_path, metric=metric)
```
aoustry: As explained above, the name _build_view is not very explicit here.

aoustry: - As explained above, the name _build_view is not very explicit here.
- Moreover view_parquet_path is not very explicit as well; in fact this parquet file is rather the metric_view after temporal aggregation

gems_views_builder/views.py
```diff
@@ -90,15 +169,43 @@ def execute(self) -> None:
...
+                metric_structure_path = metric_structure_dir / f"{metric.id}.parquet"
+                metric_structure_table.dataframe.write_parquet(
+                    metric_structure_path
+                )  # # TO DO for benchmark: test behavior when using write_parquet and not sink_parquet since metric structure table won't be heavy datum
+                # # Apply sylvan suggestions for creation of new parquet file(meta data)
+
+                filtered_lazy = pl.scan_parquet(filtered_simulation_table_path)
+                metric_structure_lazy = pl.scan_parquet(metric_structure_path)
+
+                joined_dataframe = filtered_lazy.join(
```
aoustry: Is a "lazy" join ? It should I guess, as it is pointless to store (either in disk or in RAM) the joined dataframe => I guess it is better to apply the join and groupby simultaneously (to better exploit the query engines forces)

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
...
+                    "metric_location",
+                    "breakdown_properties",
+                    "absolute_time_index",
+                    "scenario",
+                ]
+            )
+            .agg(
+                [
+                    value_agg.alias("granular_metric_value"),
+                    pl.first("granular_date").alias("granular_date"),
```
aoustry: I'm not sure to get the relevance of this line, but I might miss something!

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
...
+
+    def _build_view(self, metric_view_parquet_path: Path, metric: Metric) -> Path:
+        metric_view = pl.scan_parquet(metric_view_parquet_path)
+        time_agg = (
+            pl.col("granular_metric_value").sum()
+            if metric.time_operator == TimeOperator.SUM
+            else pl.col("granular_metric_value").mean()
+        ).alias("metric_value")
+        view_date_expr = pl.col("granular_date").alias("view_date")
+        business_view = (
```
aoustry:
```suggestion
        view = (
```

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
...
+                    "metric_value",
+                ]
+            )
         )
-        _ = filtered_simulation_table  # to avoid copilot unnecessary warnings
-        # # 2. Create metric structure table and perform SQL operations on it
-        # # Metrics are grouped by catalog, in order to prevent multiple loading of the same catalog
+        # Business view is meant to be created once, then appended to on future runs.
+        # We implement this by writing a new parquet "part" file each time.
+        dataset_dir = self.input_data_path / "views" / "metric_business_view" / metric.id
```
aoustry:
```suggestion
        dataset_dir = self.input_data_path / "views" / "metric_views" / metric.id
```

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
...
+        dataset_dir.mkdir(parents=True, exist_ok=True)
+        existing_parts = sorted(dataset_dir.glob("part-*.parquet"))
+        next_part_idx = (int(existing_parts[-1].stem.split("-")[1]) + 1) if existing_parts else 0
+        out_path = dataset_dir / f"part-{next_part_idx:05d}.parquet"
+        business_view.sink_parquet(out_path)
+        return out_path
+
+    def execute(self, cleanup_intermediate: bool = False) -> None:
```
aoustry:
```suggestion
    def build(self, cleanup_intermediate: bool = False) -> None:
```

aoustry (changes requested):
Great job. I have submitted several comments for specific lines of codes.

One major additional comments: we lack test here! At least
- One for the MetricStructureBuilder.build() -> check the content of the produced metric_structure_table
- One for the ViewBuilder.build() -> check the content of the produced Views

@dusanparipovic  you can use the small PyPSA test I sent you by Teams.

gems_views_builder/views.py
```diff
@@ -72,14 +75,90 @@ def _load_system(self, input_data_path: Path) -> InputSystem:
...
         )
-        _ = filtered_simulation_table  # to avoid copilot unnecessary warnings
-        # # 2. Create metric structure table and perform SQL operations on it
-        # # Metrics are grouped by catalog, in order to prevent multiple loading of the same catalog
+        # Business view is meant to be created once, then appended to on future runs.
+        # We implement this by writing a new parquet "part" file each time.
+        dataset_dir = self.input_data_path / "views" / "metric_views" / metric.id
+        dataset_dir.mkdir(parents=True, exist_ok=True)
+        existing_parts = sorted(dataset_dir.glob("part-*.parquet"))
+        next_part_idx = (int(existing_parts[-1].stem.split("-")[1]) + 1) if existing_parts else 0
```
tbittar: Maybe we could add some utils function to perform operations based of filenames format we manipulate to make this more readable ?

gems_views_builder/system.py
dusanparipovic: Right,done! 
Added also for models_to_components I find it useful.

gems_views_builder/system.py
dusanparipovic: Right done! 
For your 3rd bullet point I find this very useful but then I need also to pass instance of `library.yaml` e.g. in future multiple `library.yaml` instances(maybe just id(filename) and then catch it)
I'll open issue regarding this since I think in this phase we don't need it but this is something what needs to be done. 
Feel free to change [issue](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/22) description.
If you want we could implement this inside this PR.

gems_views_builder/views.py
dusanparipovic: I agree this is confusing, since component is column in SimulationTable I'll recommend that we could align on that, so we will change names of columns in MetricStructureTable.

- component_id = component 
- output_id = output

gems_views_builder/views.py
dusanparipovic: Pydantic will raise an error if user any other aggregate function e.g.(`MIN`,`MAX`,`COUNT`) or any other `val`

dusanparipovic: `polars expression` and this is passed into agg[`aggregate`]
To clear confusion: 
SUM(value) -> define in value agg but in final result it will be treated as `granular_metric_value`
Practically in SQL we write something like: 
`select SUM(value) as granular_metric_value`

gems_views_builder/views.py
dusanparipovic: Right alias removed, what `first` will do take value of first row of group whatever it is (in fact it will be same values for each row in group)

gems_views_builder/views.py
dusanparipovic: [Answer](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/20/changes#r3024828600)

gems_views_builder/views.py
dusanparipovic: Completely agree. 
- `_build_metric_view` -> `_aggregate_metric_terms`
- `_build_view` -> `_aggregate_metric_temporally`

gems_views_builder/views.py
dusanparipovic: [Answer](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/20/changes#r3024956754)

dusanparipovic: Nice comments, agree `view_parquet_path` -> `temp_metric_view`

gems_views_builder/views.py
dusanparipovic: type(`joined_dataframe`) == lazy array, what that practically mean. No real data just query execution plan which will be passed into `_aggregate_metric_terms`.
Practically how SQL always works:

- From is always firstly executed to know on which table/tables we will perform operations,so since we call sink_parquet in this function we will first perform right join from query execution plan which is passed to the function, why? 
- From declarative will look like `from filtered_simulation_table_lazy RIGHT JOIN metric_structure_lazy on filtered_simulation_table_lazy.component == metric_structure_lazy.component AND filtered_simulation_table_lazy.output == metric_structure_lazy.output` 

I would rather keep that lines out of this function, what do you think now ? 
Also I want to hear what other path of team think.

dusanparipovic: Additionally I catch that in fact we don't need to pass full metric object.

- `_aggregate_metric_terms` needs just `terms operator`
- `_aggregate_metric_temporally` needs just `time operator`

gems_views_builder/views.py
```diff
@@ -27,59 +30,190 @@
 # System -> representation of the system.yml file
 # Taxonomy -> representation of the taxonomy.yml file
 """
+EXACT_FILES = ["taxonomy.yml", "view_config.yml", "library.yml", "system.yml"]
+PREFIX_FILES = {"calendar": ".csv", "simulation_table": ".parquet"}
+TEMPORAL_METRIC_VIEW_PART = 0
```
TheoPascoli: Global mutable variable is an anti-pattern. It is not thread-safe, it is not set to 0 if you call `build()` two times. It must be a instance attribute like `self._part_counter` or even deleted. You can imagine something with the `metric_id` that is already unique

gems_views_builder/views.py
```diff
@@ -27,59 +30,190 @@
...
         self.input_data_path = input_data_path
+        # # If this function raise an error, the builder will not be able to build the views.
+        self._check_input_data_structure()
+        #
+        self.system = self._load_system()
+        self.taxonomy = load_taxonomy(self.input_data_path / "taxonomy.yml")
+        self.view_config = ViewConfig(self.input_data_path / "view_config.yml")
+        self.simulation_table = SimulationTable(
+            next(self.input_data_path.glob("simulation_table*.parquet"))
+        )  # we could have only one simulation table at this phase
```
TheoPascoli, line 58: Are we sure that we have at least one file everytime ? Like the other PR, you should handle the case where there is no file to catch the `StopIteration` that will happens

gems_views_builder/views.py
```diff
@@ -27,59 +30,190 @@
...
-        exact_files = ["taxonomy.yml", "view_config.yml"]
-        for filename in exact_files:
-            if not (input_data_path / filename).is_file():
-                raise FileNotFoundError(f"Required file '{filename}' not found in {input_data_path}")
+    def _check_number_of_required_files(self) -> None:
+        """
+        # Check if there are exactly 6 required files.
+        """
+        counter = Counter(p.is_file() for p in self.input_data_path.iterdir())
+        if counter.get(True, 0) != 6:
```
TheoPascoli: Why 6 ? This "magic number" is fragile. What happens if we add a file to the format ? You could validate each file individually by declaring a list of awaited file maybe

TheoPascoli (changes requested):
Hello Dusan, sorry for the delay. It looks good already !

I have a concern about ViewBuilder, it handles file validation, loading, joins, aggregation, writing and cleanup all in one class. The fact that tests already need `object.__new__()` to bypass `__init__` (test_views_input_checks.py:21) shows it's hard to test without the full system. I think we should split it into smaller components (loader, aggregator, writer). What do you think?

Also, open() is called without `encoding="utf-8"` in several places (catalog.py:123, library.py:92, metrics.py:93, system.py:117, taxonomy.py:65). Without it, it defaults to the system locale which is not predictable across machines.

Finally, YAML parsing like `raw["catalog"]` or `raw["view"]` will raise a bare `KeyError` if the key is missing, which is not very helpful. We should catch it and raise something explicit like "catalog.yml: missing top-level 'catalog' key".

gems_views_builder/views.py
dusanparipovic: [Input directory structure definition](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/13#issuecomment-4133189332).
This is not final version but for now input to the library will look like this. 
I agree this is kinda confusing, logic is merged into `_check_required_input_files`

gems_views_builder/views.py
dusanparipovic: See `check_input_data_structure` all check for required files and required amount of files will be done inside this function if not we will raise some error and next `ViewBuilder` object won't be created. 
Maybe you see some better approach? 
If yes I'm open to hear your thoughts ?

dusanparipovic:
> test_views_input_checks



> Hello Dusan, sorry for the delay. It looks good already !
> 
> I have a concern about ViewBuilder, it handles file validation, loading, joins, aggregation, writing and cleanup all in one class. The fact that tests already need `object.__new__()` to bypass `__init__` (test_views_input_checks.py:21) shows it's hard to test without the full system. I think we should split it into smaller components (loader, aggregator, writer). What do you think?
> 
> Also, open() is called without `encoding="utf-8"` in several places (catalog.py:123, library.py:92, metrics.py:93, system.py:117, taxonomy.py:65). Without it, it defaults to the system locale which is not predictable across machines.
> 
> Finally, YAML parsing like `raw["catalog"]` or `raw["view"]` will raise a bare `KeyError` if the key is missing, which is not very helpful. We should catch it and raise something explicit like "catalog.yml: missing top-level 'catalog' key".

Hi Theo,

I agree with your approach. It’s actually very close to how I started, keeping smaller pieces organized into separate classes and files, following SOLID principles.

However, other part of the team felt that might be too complex for an initial version. So I went along with a simpler structure to get things moving. That said, I’m absolutely open to refactoring. My original plan was already aligned with what you suggested.

I’ll prepare a proposal for restructuring and walk you through it so we can discuss and agree on the best project setup.

- Thanks for pointing out the encoding issue. That was a great catch and it’s now fixed.
- Also appreciate the note on file loading. I completely agree, and that’s been resolved as well.

gems_views_builder/metrics.py
TheoPascoli, line 80: You read this but you never use it. Same in views.py l. 177

gems_views_builder/views.py
```diff
@@ -27,59 +32,217 @@
...
+                    "metric_location",
+                    "breakdown_properties",
+                    "absolute_time_index",
+                    "scenario",
+                ]
+            )
+            .agg(
+                [
+                    value_agg.alias("granular_metric_value"),
+                    pl.first("granular_date"),  # # take a first row of group whatever it is
```
TheoPascoli: This can return null if the first column of the right join is a non match. You can do something like `pl.col("granular_date").drop_nulls().first()`, or even build it from `absolute_time_index` via a left_join with the calendar maybe

gems_views_builder/views.py
```diff
@@ -27,59 +32,217 @@
...
+            row_group_size=_PARQUET_ROW_GROUP_SIZE,
         )
-        _ = filtered_simulation_table  # to avoid copilot unnecessary warnings
-        # # 2. Create metric structure table and perform SQL operations on it
-        # # Metrics are grouped by catalog, in order to prevent multiple loading of the same catalog
+        for path in chunk_paths:
+            path.unlink(missing_ok=True)
+        return out_path
+
+    def build(self, cleanup_intermediate: bool = False) -> None:
```
TheoPascoli: After a build `simulation_table_filtered.parquet` is still in `input_data_path` and it matches `glob("simulation_table*.parquet")` at the next run. You can write intermediate files in a separate folder to prevent this. Of make `cleanup_intermediate` default to true

gems_views_builder/metrics_builder.py
```diff
@@ -58,27 +58,23 @@ def __init__(
...
+                    metric_location = self.system.get_location(component_id, term.location_ports)
+                    loc_str = metric_location if isinstance(metric_location, str) else "|".join(metric_location)
+                    rows.append(
+                        {
+                            "metric_id": self.metric.id,
+                            "component": component_id,
+                            "metric_location": loc_str,
+                            "breakdown_properties": "",
+                            "output": term.output_id,
+                            "weight_output_id": 1,
```
TheoPascoli, line 75: I think you hardcode this to 1 and never use it ?

gems_views_builder/metrics_builder.py
```diff
@@ -58,27 +58,23 @@ def __init__(
...
+                for component_id in self.system.get_instances_by_model(qualified_ref):
+                    # # locating function
+                    metric_location = self.system.get_location(component_id, term.location_ports)
+                    loc_str = metric_location if isinstance(metric_location, str) else "|".join(metric_location)
+                    rows.append(
+                        {
+                            "metric_id": self.metric.id,
+                            "component": component_id,
+                            "metric_location": loc_str,
+                            "breakdown_properties": "",
```
TheoPascoli, line 73: It is hardcoded and never used. Maybe im missing something

gems_views_builder/views.py
```diff
@@ -27,59 +32,217 @@
...
+            compression=_PARQUET_COMPRESSION,
+            compression_level=_PARQUET_COMPRESSION_LEVEL,
+            row_group_size=_PARQUET_ROW_GROUP_SIZE,
+        )
+        return out_path
+
+    def _consolidate_results(self, chunk_paths: list[Path]) -> Path:
+        results_dir = self.input_data_path / "results"
+        results_dir.mkdir(parents=True, exist_ok=True)
+        timestamp = datetime.now().strftime("%Y%m%dT%H%M%S")
```
TheoPascoli: You should specify a timezone ! UTC for exemple `datetime.now(timezone.utc)`

TheoPascoli (changes requested):
Hello Dusan, nice work for now, still some room for improvements !

`build()` at `views.py:283` uses `how="right"`, so a component in `metric_structure` but absent from the simulation becomes a row with `value = null`.

Then `_aggregate_metric_terms` (`views.py:136`) sums it, and Polars `sum()` over only null values returns `0.0`, not `null`:

```python
pl.DataFrame({"k": ["a"], "v": [None]}).group_by("k").agg(pl.col("v").sum())
# → 0.0
```

Result: missing components show up as phantom `0.0` rows in the final parquet, indistinguishable from real zeros. Silent. Only affects `TermsOperator.SUM`.

Fix: filter before aggregating:

```python
joined.filter(pl.col("value").is_not_null())
```
-----------------------
Also, 
The comment on `_aggregate_metric_temporally` (`views.py:210`) says the view is **"appended to on future runs"** via part files.

Can you confirm this actually works?

From reading the code, `_consolidate_results` deletes the parts at `views.py:237`, and `_part_counter` resets to `0` in `__init__`, so I don’t see where the append happens.

Suggested test:

```python
ViewBuilder(path).build()  # run 1
ViewBuilder(path).build()  # run 2

# Does the second result contain run 1's rows?
# I expect no.
```
Tell me if you have any questions !

gems_views_builder/metrics_builder.py
aoustry: In fact `weight_output` is not really used yet, but will be in future workflow to compute weighted average and weighted sums in the groupby (term-operators).

gems_views_builder/metrics_builder.py
aoustry: You mean `breakdown_properties` ? It will be used in future workflows...

aoustry (dismissed):

gems_views_builder/metrics_builder.py
dusanparipovic: `weight_output_id` &&  `breakdown_properties` aren't used currently, so that's reason why it's hardcoded. 

- `breakdown_properties` isn't defined for now they will be define inside `catalo.yml` files. 
- `weight_output_id` also it's mentioned that will be ignored in first version(Business View Building Process pdf)
If you prefer not to have hardcoded values I could comment this 2 lines , but just keep in source code just to know what to update in future.

gems_views_builder/metrics_builder.py
dusanparipovic: [Answer](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/20/changes#r3085800401)

gems_views_builder/metrics.py
dusanparipovic: Nice catch! 
This isn't used currently but I figured out fact that I need to add more enumerators. (`hour`, `day`, `week`, `month`, `year`)

aoustry (approved):

TheoPascoli (approved):

PR #31, opened 2026-04-26
refactor(view builder): decomposition of code into smaller dedicated pieces

dusanparipovic:
The PR introduces:

- `common.py` - file for common things used over multiple files inside project, currently only meta data for parquet files
- `InputValidator` - dedicated class which is engaged for validation of input files
- `Loader` - class engaged loading of all input data 
- `Aggregator` - class engaged for `aggregate metric terms`, and `aggregate metric temporally`
- `Writer` - class engaged for consolidation of results and writing metric structure table
- `tests_aggregator` 
- `tests_loader` 
- `tests_writer`   
Refactor:
- Refactored tests to use dedicated functions/classes in order to achieve testing each piece of code isolated 
- Utilized new objects and functions inside `views.py`


gems_views_builder/input_validator.py
```diff
@@ -0,0 +1,74 @@
...
+        for prefix, expected_suffix in PREFIX_FILES.items():
+            match = next(self.input_data_path.glob(f"{prefix}*"), None)
+            if match is None:
+                raise FileNotFoundError(f"Required file starting with '{prefix}' not found in {self.input_data_path}")
+            if match.suffix != expected_suffix:
+                raise ValueError(f"File '{match.name}' starting with '{prefix}' must be a '{expected_suffix}' file")
+            files_counter[match.name] += 1
+
+        # # Check counter
+        if sum(files_counter.values()) != len(EXACT_FILES) + len(PREFIX_FILES):
```
TheoPascoli: I cant see a scenario where this code can be reached. It looks like you raise in all the cases above. Maybe I'm wrong

gems_views_builder/loader.py
```diff
@@ -0,0 +1,26 @@
...
+
+        self.system = self._load_system()
+        self.taxonomy = load_taxonomy(self.input_data_path / "taxonomy.yml")
+        self.view_config = ViewConfig(self.input_data_path / "view_config.yml")
+        self.simulation_table = SimulationTable(
+            next(self.input_data_path.glob("simulation_table*.parquet"))
+        )  # # we could have only one simulation table at this phase of development
+        self.model_library = ModelLibrary(
+            self.input_data_path / "library.yml"
+        )  # # must be named like this for now, in future when we enable user to have more than one libraries we should decide pattern to use
```
TheoPascoli, line 29: This init performs a lot of actions, loading yaml, scanning parquets... class constructor shouldn't do I/O. I suggest you define a load method to separate class creation from I/O

gems_views_builder/writer.py
```diff
@@ -0,0 +1,43 @@
+from datetime import datetime, timezone
+from pathlib import Path
+
+import polars as pl
+
+from gems_views_builder.common import PARQUET_COMPRESSION, PARQUET_COMPRESSION_LEVEL, PARQUET_ROW_GROUP_SIZE
+from gems_views_builder.metrics_builder import MetricStructureTable
```
TheoPascoli: Importing `MetricStructureTable` from `metrics_builder` creates a reverse dependecy (low-level module import business module). You use this import only for one attribute, you could just pass it

gems_views_builder/views.py
```diff
@@ -47,202 +36,36 @@ def __init__(
...
-        # # 1. Filter simulation table (written to disk)
+        # # 1. Validate input data, if it fails, the builder will not be able to build the views.
+        InputValidator(self.input_data_path).validate()
+        # # 2. Load input data, if it fails, the builder will not be able to build the views.
+        loader = Loader(self.input_data_path)
+        self.system = loader.system
+        self.taxonomy = loader.taxonomy
+        self.view_config = loader.view_config
+        self.simulation_table = loader.simulation_table
+        self.model_library = loader.model_library
```
TheoPascoli: You can just have `self.loader`, and use it whenever you need ont of the five variables

TheoPascoli (changes requested):
Looks better now ! 
You could add more tests for the Loader and also the Aggregator which contains a lot of computation logic

gems_views_builder/input_validator.py
dusanparipovic: You're right sorry for this.

TheoPascoli (dismissed):
Overall the refactor is a solid step forward. Good job! 
Just worth keeping in mind that test coverage stays fairly minimal, especially on the Aggregator which holds most of the computation logic. Could be worth a follow-up.

gems_views_builder/aggregator.py
```diff
@@ -0,0 +1,110 @@
...
+        self._part_counter = 0
+
+    def aggregate_metric_terms(
+        self, joined_dataframe: pl.LazyFrame, metric_term_operator: TermsOperator, metric_id: str
+    ) -> Path:
+        """
+        2b step from POC
+        2b-1 Right join TIME_FILTERED_SIMULATION_TABLE with METRIC_STRUCTURE_TABLE on component and output
+        2b-2 Group by metric_id, metric_location, breakdown_properties, absolute_time_index, scenario
+        """
```
tbittar: The description does not exactly match what happens inside the function. From the cod, I would say it is only step 2b-2

gems_views_builder/aggregator.py
```diff
@@ -0,0 +1,110 @@
...
+            compression=PARQUET_COMPRESSION,
+            compression_level=PARQUET_COMPRESSION_LEVEL,
+            row_group_size=PARQUET_ROW_GROUP_SIZE,
+        )
+        return out_path
+
+    def aggregate_metric_temporally(
+        self, metric_view_parquet_path: Path, metric_time_operator: TimeOperator, metric_id: str
+    ) -> Path:
+        metric_view = pl.scan_parquet(metric_view_parquet_path)
```
tbittar, line 78: This means that in the workflow we write and then read the metric_view file ?

gems_views_builder/aggregator.py
```diff
@@ -0,0 +1,110 @@
...
+
+import polars as pl
+
+from gems_views_builder.catalog import TermsOperator, TimeOperator
+from gems_views_builder.common import PARQUET_COMPRESSION, PARQUET_COMPRESSION_LEVEL, PARQUET_ROW_GROUP_SIZE
+
+
+class Aggregator:
+    def __init__(self, input_data_path: Path) -> None:
+        self.input_data_path = input_data_path
```
tbittar: It seems that this attribute is used to write output parquet files
```suggestion
        self.output_data_path = output_data_path
```

gems_views_builder/aggregator.py
```diff
@@ -0,0 +1,110 @@
...
+                    "view_date",
+                    "scenario",
+                    "metric_value",
+                ]
+            )
+        )
+        # Business view is meant to be created once, then appended to on future runs.
+        # We implement this by writing a new parquet "part" file each time.
+        dataset_dir = self.input_data_path / "temporal_aggregation"
+        dataset_dir.mkdir(parents=True, exist_ok=True)
```
tbittar: Is it relevant to do this only once in the class constructor rather than each time the aggregate function is called ? @TheoPascoli

gems_views_builder/aggregator.py
```diff
@@ -0,0 +1,110 @@
...
+                    "breakdown_properties",
+                    "absolute_time_index",
+                    "scenario",
+                    "granular_metric_value",
+                    "granular_date",
+                ]
+            )
+        )
+        out_dir = self.input_data_path / "views" / "metric_view"
+        out_dir.mkdir(parents=True, exist_ok=True)
```
tbittar: Is it relevant to do this only once in the class constructor rather than each time the aggregate function is called ? @TheoPascoli

gems_views_builder/input_validator.py
```diff
@@ -0,0 +1,70 @@
+from pathlib import Path
+
+EXACT_FILES = ["taxonomy.yml", "view_config.yml", "library.yml", "system.yml"]
+PREFIX_FILES = {"calendar": ".csv", "simulation_table": ".parquet"}
+
+
+class InputValidator:
+    """
+    Validator of the input data
```
tbittar: Either add more precision on the expected input data (you can simply take the description of the validate function and put it here for better visibility) or remove this comment (as by itself it does not give much information)

gems_views_builder/input_validator.py
```diff
@@ -0,0 +1,70 @@
...
+    def validate(self) -> None:
+        """
+        Expected files:
+        - taxonomy.yml
+        - view_config.yml
+        - library.yml
+        - system.yml
+        - simulation_table.parquet
+        - calendar.csv
+        - catalogs directory with 1 * catalogs without strict name convention for now
```
tbittar: Why 1 * catalogs ? Do you mean at least 1 catalog ?

gems_views_builder/views.py
```diff
@@ -289,23 +101,19 @@ def build(self, cleanup_intermediate: bool = True) -> None:
                     how="right",
                 )
 
-                metric_view_parquet_path = self._aggregate_metric_terms(
+                metric_view_parquet_path = self.aggregator.aggregate_metric_terms(
                     joined_dataframe=joined_dataframe, metric_term_operator=metric.terms_operator, metric_id=metric.id
                 )
-                temp_metric_view = self._aggregate_metric_temporally(
+                temp_metric_view = self.aggregator.aggregate_metric_temporally(
```
tbittar, line 107: In the process, we write a lot of intermediate files to disk (filtered simulation table, metric_table with term agregation, metric table with terms and time agregation...) 

Isn't it better (and possible) to have a workflow where we go from the simulation table + metric table directly to the final metric view directly in memory (or part of it if we need to write by chuncks) ? @aoustry @TheoPascoli

tests/test_aggregator.py
```diff
@@ -0,0 +1,59 @@
...
+    ).lazy()
+
+
+def test_aggregate_metric_terms_sum(tmp_path: Path) -> None:
+    agg = Aggregator(tmp_path)
+    out_path = agg.aggregate_metric_terms(_joined_df([2.0, 3.0]), TermsOperator.SUM, metric_id="M")
+
+    df = pl.read_parquet(out_path)
+    assert df.shape[0] == 1
+    assert df["granular_metric_value"][0] == 5.0
```
tbittar: use pytest.approx

tests/test_aggregator.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_aggregate_metric_terms_avg(tmp_path: Path) -> None:
+    agg = Aggregator(tmp_path)
+    out_path = agg.aggregate_metric_terms(_joined_df([2.0, 3.0]), TermsOperator.AVG, metric_id="M")
+
+    df = pl.read_parquet(out_path)
+    assert df.shape[0] == 1
+    assert df["granular_metric_value"][0] == 2.5
```
tbittar: same

tests/test_aggregator.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_aggregate_metric_temporally_sum_and_part_counter(tmp_path: Path) -> None:
+    agg = Aggregator(tmp_path)
+
+    metric_view_path = agg.aggregate_metric_terms(_joined_df([1.0, 2.0]), TermsOperator.SUM, metric_id="M")
+
+    part0 = agg.aggregate_metric_temporally(metric_view_path, TimeOperator.SUM, metric_id="M")
+    df0 = pl.read_parquet(part0)
+    assert df0.shape[0] == 1
+    assert df0["metric_value"][0] == 3.0
```
tbittar: same

gems_views_builder/input_validator.py
dusanparipovic: Yes right, in future when we made complex tests we could reference more than one catalog file.
It's just regex shortcut for writing

gems_views_builder/aggregator.py
dusanparipovic: right nice catch

gems_views_builder/aggregator.py
dusanparipovic: Right, this is copy paste miss

gems_views_builder/views.py
dusanparipovic: Main reason why we're doing this is fact we can't do .append() for `parquet` type of files as we do for `csv`.
Maybe there is some way, for that I suggest monitor of memory peak for this part of aggregation. 
Just one remark, we're doing something very similar for APG(on antares legacy models) and they are fine if they got results into chunks in fact that results will be imported to `microsoft power bi` or `databricks`.
Maybe we could consider to remove merging part ?

gems_views_builder/aggregator.py
dusanparipovic: Yes, right.
Line 67 will say you everything, we're doing this because then we want to perform operation on disk.(e.g. on lazy array)

gems_views_builder/aggregator.py
dusanparipovic: I agree with Thomas, we could move this part to constructor. 
Ofc if Theo agree.

gems_views_builder/aggregator.py
dusanparipovic: [Same answer.](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/31#discussion_r3181759207)

gems_views_builder/aggregator.py
TheoPascoli: Yes you're right nice catch

TheoPascoli (approved):

tbittar (approved):

PR #35, opened 2026-05-01
feat(filtering): add support for filter and breakdown property 

dusanparipovic:


requirements.txt
```diff
@@ -6,4 +6,4 @@ pyyaml==6.0.2
 # Pydantic
 pydantic>=2.11.7
 # gemspy
-gemspy>=0.0.4
\ No newline at end of file
+gemspy @ git+https://github.com/AntaresSimulatorTeam/GemsPy.git@properties
```
dusanparipovic: This will be removed once we release new version of gemspy library

gems_views_builder/catalog.py
```diff
@@ -43,13 +43,33 @@ class TermData(ViewBuilderBasedModel):
...
+    key: str
+    value: str
+
+
+class BreakdownKey(ViewBuilderBasedModel):
+    key: str
+
+
+class BreakdownProperty(ViewBuilderBasedModel):
+    """Legacy single breakdown clause (key,value)."""
```
aoustry: What is "Legacy" ? Your first trial? :D

gems_views_builder/catalog.py
```diff
@@ -43,13 +43,33 @@ class TermData(ViewBuilderBasedModel):
...
     id: str
     terms: list[TermData]
     terms_operator: TermsOperator
     time_operator: TimeOperator
-    breakdown_property: str | None = None
-    filter: tuple[str, str] | None = None
+    # New schema: group-by keys
+    breakdown: list[BreakdownKey] | None = None
+    # Legacy schema: single clause
+    breakdown_property: BreakdownProperty | None = None
```
aoustry: I guess you should delete BreakdownProperty  here, no ? And also, please remove the comments (New/Legacy schema)

gems_views_builder/catalog.py
aoustry: I guess you should delete BreakdownProperty

gems_views_builder/catalog.py
```diff
@@ -43,13 +43,33 @@ class TermData(ViewBuilderBasedModel):
     weight_output_id: str | None = None
 
 
+class PropertyFilter(ViewBuilderBasedModel):
+    key: str
+    value: str
+
+
+class BreakdownKey(ViewBuilderBasedModel):
```
aoustry: Rename PropertyKey?

gems_views_builder/catalog.py
```diff
@@ -43,13 +43,33 @@ class TermData(ViewBuilderBasedModel):
...
+    key: str
+
+
+class BreakdownProperty(ViewBuilderBasedModel):
+    """Legacy single breakdown clause (key,value)."""
+
+    key: str
```
aoustry, line 51: Use the class "PropertyKey" ?

gems_views_builder/catalog.py
```diff
@@ -99,13 +119,31 @@ def _to_term(term_data: TermData) -> Term:
 
 
 def _to_metric(metric_data: MetricData) -> Metric:
+    filter_clauses: tuple[tuple[str, str], ...] | None
+    if metric_data.filter is None:
+        filter_clauses = None
+    elif isinstance(metric_data.filter, list):
+        filter_clauses = tuple((f.key, f.value) for f in metric_data.filter)
+    else:
+        filter_clauses = ((metric_data.filter.key, metric_data.filter.value),)
```
aoustry: What is the reason of this distinction ? looks weird

gems_views_builder/catalog.py
```diff
@@ -99,13 +119,31 @@ def _to_term(term_data: TermData) -> Term:
...
+    breakdown: tuple[str, ...] | None = None
+    if metric_data.breakdown is not None:
+        breakdown = tuple(b.key for b in metric_data.breakdown)
+    elif metric_data.breakdown_property is not None:
+        # Legacy: treat breakdown_property as a filter clause on that key and
+        # use the key itself as breakdown dimension.
+        breakdown = (metric_data.breakdown_property.key,)
+        filter_clauses = tuple(
+            list(filter_clauses or ()) + [(metric_data.breakdown_property.key, metric_data.breakdown_property.value)]
+        )
```
aoustry: I guess this part should be deleted

gems_views_builder/metrics_builder.py
```diff
@@ -38,9 +40,54 @@ class MetricStructureTable:
...
 
-class MetricStructureBuilder:
-    """Build metric structure rows without loading unrelated datasets."""
+def _component_properties(component: Component) -> dict[str, str]:
+    """
+    Normalize `component.properties` to a dict[str,str].
+
+    Depending on Gems parser versions, properties may be:
+    - dict-like {key: value}
+    - list-like of items with (id|key, value) fields
```
aoustry: Why these two options??

gems_views_builder/system.py
```diff
@@ -10,74 +10,98 @@
...
-            groups[model_ref].append(component.id)
+            groups[key].append(component.id)
         return groups
 
     def build_component_port_connections(self) -> dict[tuple[str, str], set[str]]:
         """
         Iterate over connections and for each component and port add other side of connection in dictionary
         """
+
+        def _endpoint(conn: Any, idx: int) -> tuple[str, str] | None:
```
aoustry, line 80: Is there a reason why you modify this method build_component_port_connections in this PR ? This does not seem related to the main topic...

gems_views_builder/catalog.py
```diff
@@ -43,13 +43,33 @@ class TermData(ViewBuilderBasedModel):
     weight_output_id: str | None = None
 
 
+class PropertyFilter(ViewBuilderBasedModel):
+    key: str
```
aoustry: PropertyKey (cf below?)

gems_views_builder/catalog.py
```diff
@@ -43,13 +43,33 @@ class TermData(ViewBuilderBasedModel):
     weight_output_id: str | None = None
 
 
+class PropertyFilter(ViewBuilderBasedModel):
```
aoustry: PropertyTuple ?

gems_views_builder/catalog.py
```diff
@@ -77,8 +97,8 @@ class Metric:
     terms: list[Term]
     terms_operator: TermsOperator
     time_operator: TimeOperator
-    breakdown_property: str | None = None
-    filter: tuple[str, str] | None = None
+    breakdown: tuple[str, ...] | None = None
+    filter: tuple[tuple[str, str], ...] | None = None
```
aoustry: In the signature, yse the abstract classes that you defined before.

gems_views_builder/metrics_builder.py
```diff
@@ -38,9 +40,54 @@ class MetricStructureTable:
...
+
+def _component_matches_property_filter(component: Component, clauses: tuple[tuple[str, str], ...] | None) -> bool:
+    """If the metric defines filters, the component must match all (key,value) clauses."""
+    if clauses is None:
+        return True
+    props = _component_properties(component)
+    return all(props.get(k) == v for k, v in clauses)
+
 
+def _format_breakdown_properties(props: dict[str, str], keys: tuple[str, ...] | None) -> str:
```
aoustry: - please clarify in the var names that props are properties of a component, whereas keys correspond to the breakdown keys of the metric.
- Do we agree that this function should not be designed to handle objects deriving from the BreakdownProperty class ? As explained in the other comments, the BreakdownProperty class should be deleted. 
- Why not using, in the signature of the method, the abstract dataclasses that you created? (PropertyKey... etc)

tests/test_filtering_and_breakdown.py
```diff
@@ -0,0 +1,85 @@
+import shutil
+from pathlib import Path
+
+import polars as pl
+import pytest
+
+from gems_views_builder.views import ViewBuilder
+
```
aoustry, line 26: I guess we will need more tests, for instance precise evaluation of the expected production level for a given technology and for a given company.

aoustry (changes requested):
Thank you Dusan!

Here is a first batch of comments! Please take them into account (+ unzip the test folder please), and I will then re-review the code more extensively.

Thanks!

gems_views_builder/system.py
dusanparipovic: No I just realized while integrating new version of GemsPy that this function could be quickly refactored for more readability. 
If previous version fits better for you I could revert changes.

gems_views_builder/catalog.py
dusanparipovic: I did simplification of function. 

- In fact filter could be just one key,value pair 
- Breakdown could be listed as multiple keys

gems_views_builder/metrics_builder.py
dusanparipovic: This was here because I'm waiting approval on this [PR](https://github.com/AntaresSimulatorTeam/GemsPy/pull/205/changes)

requirements.txt
dusanparipovic: [PR](https://github.com/AntaresSimulatorTeam/GemsPy/pull/205)

pyproject.toml
```diff
@@ -12,7 +12,8 @@ dependencies = [
     "pyarrow>=14.0.0",
     "pyyaml==6.0.2",
     "pydantic>=2.11.7",
-    "gemspy>=0.0.4",
+    # GemsPy (PR #205 / `properties` branch): installs the `gems` top-level package.
+    "gemspy @ git+https://github.com/AntaresSimulatorTeam/GemsPy.git@properties",
```
dusanparipovic, line 16: Note for me: This needs to be updated before merging

gems_views_builder/system.py
aoustry: It was just to know, thank you for the improvement

gems_views_builder/catalog.py
```diff
@@ -43,13 +44,28 @@ class TermData(ViewBuilderBasedModel):
     weight_output_id: str | None = None
 
 
+class PropertyTuple(ViewBuilderBasedModel):
```
aoustry: Sorry for the additional suggestion, I think that "PropertySchema" could better match the existing namings in GemsPy.

gems_views_builder/catalog.py
aoustry: I think this comment is still valid, no ?

gems_views_builder/metrics_builder.py
```diff
@@ -38,9 +39,27 @@ class MetricStructureTable:
...
+) -> str:
+    if not breakdown_keys:
+        return "{}"
+    pairs: list[str] = []
+    for breakdown_key in breakdown_keys:
+        key = breakdown_key.key
+        if key not in component_properties:
+            return "{}"
+        pairs.append(f"({key},{component_properties[key]})")
+    return "{" + ",".join(pairs) + "}"
```
aoustry, line 62: Don't we need a sorting step, to be sure that the order of declaration of properties does not matter ?

aoustry (commented):
That's great! Just a few additional comments + we need at least one additional test that aims at checking that the order of declaration of properties has no importance for the results.

gems_views_builder/metrics_builder.py
```diff
@@ -38,9 +39,27 @@ class MetricStructureTable:
...
+def _format_breakdown_properties(
+    component_properties: dict[str, str], breakdown_keys: tuple[PropertyKey, ...] | None
+) -> str:
+    if not breakdown_keys:
+        return "{}"
+    pairs: list[str] = []
+    for breakdown_key in breakdown_keys:
+        key = breakdown_key.key
+        if key not in component_properties:
+            return "{}"
```
aoustry:
```suggestion
            pairs.append(f"({key},None)")
```

aoustry (commented):
Please also add tests to check the behavior of breakdown when properties key are not present in the components (see my suggestion on line 57 of metrics_builder.py).

I would suggest that the behaviour should be { (country, None), (company, rhonepower), (technology, None)} if country and technology are not declared for a given component.

gems_views_builder/metrics_builder.py
dusanparipovic: ### system.yml - same values, different list order

**gen_a** (technology first):
```yaml
properties:
  - id: technology
    value: gas
  - id: company
    value: rhonepower
```

**gen_b** (company first):
```yaml
properties:
  - id: company
    value: rhonepower
  - id: technology
    value: gas
```

After load, both components have the same map:
`{technology: gas, company: rhonepower}`

YAML list order is not used when building the breakdown string.

---

**catalog.yml** - order used for every component:
```yaml
breakdown:
  - key: technology
  - key: company
```

**What the code does:**

1. For each catalog breakdown key, look up that key in `component.properties`.
2. Build pairs in catalog order - `(technology, gas)`, `(company, rhonepower)`.
3. Join - `{(technology,gas),(company,rhonepower)}` - same string for gen_a and gen_b.
4. Aggregator `groupby` on `breakdown_properties` - one bucket - both generators summed.

gems_views_builder/catalog.py
dusanparipovic: This two things represents different part of catalog I would rather separate it, maybe we could name it BreakdownKey ? 
```
Example:  
breakdown:
  - key: technology 
  - key: company 
  - ....
  - key: something 
```

```
filter:
    - key: technology
      value: nuclear
```

gems_views_builder/metrics_builder.py
aoustry: Understood, thanks!

aoustry: @dusanparipovic  But still, we need a test that indeed checks this.

gems_views_builder/catalog.py
aoustry: But they represent same parts from the System! In both cases, these are keys of **Properties**, and these keys can be used at several places of the catalog. I think that the nature of this dataclass should not depend on the usecase, but on the essence...

gems_views_builder/metrics_builder.py
```diff
@@ -56,25 +77,51 @@ def __init__(
...
+                    component = self.system.get_component(component_id)
+
+                    # # Decide does the component matches the filter, if yes they will contribute to the metric
+                    if _check_filter_matches(component, self.metric.filter):
+                        metric_location = self.system.get_location(component_id, term.location_ports)
+                        locations = [metric_location] if isinstance(metric_location, str) else list(metric_location)
+
+                        breakdown_properties = _format_breakdown_properties(component.properties, self.metric.breakdown)
+
+                        for location in locations:
```
dusanparipovic: This part of code is extended logic and bug fix for get_location.
In fact user can pass: 

- `None` - we will put component itself if location_port is `None` 
- `one port e.g. str` - we will return one or more location components(from other side of connection) and that will be separated into multiple rows. 
Open question: Do we wan't to split that into multiple rows(that's correct from my perspective) 
- `tuple ports e.g. multiple ports` - we will return all components from other side of connection. 
**Remark** we don't care if through which port is connected our component to other side because anyways we will separate them into multiple rows.
**Remark num2**: 
Imagine scenario where we have portA,portB mentioned in metric and both of them makes connection to same component, so basically our function will return 2x same component and that will represent 2 separated frows in our metric structure. 
Feel free to correct me if I'm wrong but I think I did well assumption. 
**Remark num3** : all of that scenarios tested. 
@aoustry

PR #49, opened 2026-06-04
feat(logging): add pipeline logger

### LINKED ISSUE #32: Logging
- Add detailed logging to track the script’s current execution point.
- This is especially useful for large studies involving multiple catalogs, as it provides better visibility into where the process is at any given time.
- Refactor the implementation to ensure the logging is consistent, clear, and easy to follow.

tbittar:
## Summary

- Add `PipelineLogger` class in `common.py`: wraps `logging.Logger` with per-handler `ContextFilter` labels and a file-handler lifecycle (`configure_pipeline_run` / `close_pipeline_run`)
- Instrument all modules with `logger.info` calls so every load, parse, aggregation step, and pipeline phase is traceable
- Each pipeline run writes a timestamped log file to `<study>/logs/pipeline-<timestamp>.log`
- `ViewBuilder.build()` wrapped in `try/finally` to guarantee `close_pipeline_run()` even on failure

## Test plan

- [x] Run an existing test (`test_view_builder.py`) — log messages appear on stdout with context labels
- [x] Confirm a `logs/` directory and timestamped log file are created in the test dataset folder
- [x] Confirm no log file is leaked if the pipeline raises an exception


tbittar (commented):

gems_views_builder/common.py
```diff
@@ -1,5 +1,114 @@
...
+from typing import Iterator, Literal
 
 PARQUET_COMPRESSION: Literal["zstd"] = "zstd"
 PARQUET_COMPRESSION_LEVEL = 3
 PARQUET_ROW_GROUP_SIZE = 64_000
+LOG_FORMAT = "[%(asctime)s][%(levelname)s][%(context)s] %(message)s"
+DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+
+class ContextFilter(logging.Filter):
```
tbittar: **Thread safety**: `ContextFilter.context` is a plain attribute mutated across all handlers on the shared `logger` singleton. `pytest-xdist` (already in dev deps) runs tests in parallel workers that share the same process, so context labels from one test can bleed into another. Use `threading.local` inside `ContextFilter`, or a per-call-site `logging.LoggerAdapter`, to keep contexts isolated.

gems_views_builder/common.py
```diff
@@ -1,5 +1,114 @@
...
+    def close_pipeline_run(self) -> None:
+        if self._pipeline_handler is None:
+            return
+        self.logger.removeHandler(self._pipeline_handler)
+        self._pipeline_handler.close()
+        self._pipeline_handler = None
+        self._pipeline_log_path = None
+        self.set_context(self._default_context)
+
+    def info(self, msg: str) -> None:
```
tbittar: **Wrapper vs standard interface**: `PipelineLogger` wraps `logging.Logger` and re-exposes `info`/`warning`/`error`/`exception` as thin forwarding methods, duplicating the stdlib API. Callers cannot use `debug`, `critical`, or `log` without adding more boilerplate. Consider a `logging.LoggerAdapter` instead — it composes cleanly with the standard interface and the `extra` dict is the natural home for the context label.

gems_views_builder/catalog.py
```diff
@@ -110,24 +111,33 @@ def _to_metric(metric_data: MetricData) -> Metric:
...
         metrics={metric.id: _to_metric(metric) for metric in parsed.metrics_definition},
     )
+    logger.info(
+        f"Catalog {catalog.id!r} loaded with taxonomy {catalog.taxonomy!r} and {len(catalog.metrics)} metric(s)"
+    )
+    return catalog
 
 
 def _load_catalog_file(catalog_file_path: Path) -> CatalogData:
+    logger.info(f"Parsing catalog YAML from {catalog_file_path}")
```
tbittar: **Logging verbosity**: Calls like `"Parsing catalog YAML from ..."`, `"Catalog YAML parsed successfully"`, `"Looking up metric X in catalog Y"`, and `"Metric X found"` are internal bookkeeping. They should be `logger.debug(...)` not `logger.info(...)`. Step-level summaries (`"Step 1: Filtering..."`, `"Pipeline complete"`, `"Catalog X loaded with N metrics"`) are appropriate at INFO; per-file parse confirmations are not.

tbittar: debug

gems_views_builder/catalog.py
```diff
@@ -110,24 +111,33 @@ def _to_metric(metric_data: MetricData) -> Metric:
...
     with open(catalog_file_path, encoding="utf-8") as f:
         raw = yaml.safe_load(f)
     if "catalog" not in raw:
         raise ValueError(f"catalog.yml file {catalog_file_path} is missing the 'catalog' key at the root")
+    logger.info(f"Catalog YAML parsed successfully from {catalog_file_path}")
```
tbittar: debug

gems_views_builder/catalog.py
```diff
@@ -110,24 +111,33 @@ def _to_metric(metric_data: MetricData) -> Metric:
...
     return CatalogData.model_validate(raw["catalog"])
 
 
 def get_catalog_metric(catalog: Catalog, metric_id: str) -> Metric:
+    logger.info(f"Looking up metric {metric_id!r} in catalog {catalog.id!r}")
```
tbittar: debug

gems_views_builder/catalog.py
```diff
@@ -110,24 +111,33 @@ def _to_metric(metric_data: MetricData) -> Metric:
...
     if metric_id not in catalog.metrics:
         raise ValueError(f"Metric {metric_id} not found in catalog {catalog.id}")
+    logger.info(f"Metric {metric_id!r} found in catalog {catalog.id!r}")
```
tbittar: debug

gems_views_builder/library.py
```diff
@@ -94,23 +95,33 @@ def load(cls, library_file_path: Path) -> "ModelLibrary":
...
 
     def load_into_self(self) -> "ModelLibrary":
+        logger.info(f"Loading model library from {self.file}")
         parsed = self._load_library_file(self.file)
         self.id = parsed.id
         self.description = parsed.description or ""
         self.port_types = parsed.port_types
         self.models = {m.id: m for m in parsed.models}
+        logger.info(
+            f"Library schema loaded: id={self.id!r}, {len(self.port_types)} port type(s), {len(self.models)} model(s)"
```
tbittar:
```suggestion
            f"Library {self.id!r} loaded, containing {len(self.port_types)} port type(s),{len(self.models)} model(s)"
```

gems_views_builder/library.py
```diff
@@ -94,23 +95,33 @@ def load(cls, library_file_path: Path) -> "ModelLibrary":
...
+        )
         self.models_by_taxonomy_category = {}
         for m in parsed.models:
             if not m.taxonomy_category:
                 continue
             self.models_by_taxonomy_category.setdefault(m.taxonomy_category, []).append(m.id)
+        logger.info(
```
tbittar: debug

gems_views_builder/library.py
```diff
@@ -94,23 +95,33 @@ def load(cls, library_file_path: Path) -> "ModelLibrary":
...
+            f"Library indexing complete: {len(self.models_by_taxonomy_category)} taxonomy categor"
+            f"{'y' if len(self.models_by_taxonomy_category) == 1 else 'ies'}"
+        )
         return self
 
     def _load_library_file(self, library_file_path: Path) -> LibraryData:
+        logger.info(f"Parsing library YAML from {library_file_path}")
```
tbittar: debug

gems_views_builder/library.py
```diff
@@ -94,23 +95,33 @@ def load(cls, library_file_path: Path) -> "ModelLibrary":
...
         with open(library_file_path, encoding="utf-8") as f:
             raw = yaml.safe_load(f)
         if "library" not in raw:
             raise ValueError(f"library.yml file {library_file_path} is missing the 'library' key at the root")
+        logger.info("Library YAML parsed successfully")
```
tbittar: debug

gems_views_builder/loader.py
```diff
@@ -18,15 +19,21 @@ def load(cls, input_data_path: Path) -> "Loader":
 
     def load_into_self(self) -> "Loader":
         """Perform all input data I/O and populate attributes."""
+        logger.info(f"Loading inputs from {self.input_data_path}")
         self.system = self._load_system()
```
tbittar: Add a log that the system is loaded

gems_views_builder/metrics.py
```diff
@@ -98,9 +100,14 @@ def load_into_self(self) -> "ViewConfig":
...
         # Internal helper: grouped by catalog
         self.catalog_to_metrics = self._group_metrics_by_catalog(parsed.metrics)
+        logger.info(
+            f"View config {self.id!r} loaded: calendar={self.calendar_id!r}, "
+            f"catalogs={len(self.catalog_ids)}, metric groups={len(self.catalog_to_metrics)}"
+        )
         return self
 
     def _group_metrics_by_catalog(self, metrics: list[MetricRef]) -> dict[str, list[str]]:
+        logger.info(f"Grouping {len(metrics)} metric reference(s) by catalog")
```
tbittar: debug

gems_views_builder/metrics.py
```diff
@@ -110,13 +117,16 @@ def _group_metrics_by_catalog(self, metrics: list[MetricRef]) -> dict[str, list[
                 )
             catalog_id, metric_id = metric.id.split(".", 1)
             catalog_to_metrics[catalog_id].append(metric_id)
+            logger.info(f"Mapped metric reference {metric.id!r} to catalog {catalog_id!r}")
```
tbittar: debug

gems_views_builder/metrics_builder.py
```diff
@@ -56,12 +57,24 @@ def __init__(
         self.model_library = model_library
 
     def build(self) -> MetricStructureTable:
+        logger.info(f"[{self.metric.id}] Building metric structure table ({len(self.metric.terms)} term(s))")
         rows: list[dict[str, object]] = []
         for term in self.metric.terms:
+            logger.info(
```
tbittar: debug

gems_views_builder/metrics_builder.py
```diff
@@ -56,12 +57,24 @@ def __init__(
...
+                f"[{self.metric.id}] Processing term for taxonomy category {term.taxonomy_category!r} "
+                f"and output {term.output_id!r}"
+            )
             model_ids = self.model_library.get_components_in_taxonomy_category(term.taxonomy_category)
+            logger.info(
```
tbittar: debug

gems_views_builder/metrics_builder.py
```diff
@@ -56,12 +57,24 @@ def __init__(
...
+                f"[{self.metric.id}] Found {len(model_ids)} model(s) in taxonomy category {term.taxonomy_category!r}"
+            )
             for model_id in model_ids:
                 qualified_ref = f"{self.model_library.id}.{model_id}"
-                for component_id in self.system.get_instances_by_model(qualified_ref):
+                component_ids = self.system.get_instances_by_model(qualified_ref)
+                logger.info(
```
tbittar: debug

gems_views_builder/taxonomy.py
```diff
@@ -54,17 +55,22 @@ class Taxonomy:
...
         id=parsed.id,
         description=parsed.description,
         categories=parsed.categories,
     )
+    logger.info(f"Taxonomy {taxonomy.id!r} loaded with {len(taxonomy.categories)} categor(ies)")
+    return taxonomy
 
 
 def _load_taxonomy_file(taxonomy_file_path: Path) -> TaxonomyData:
+    logger.info(f"Parsing taxonomy YAML from {taxonomy_file_path}")
```
tbittar: debug

gems_views_builder/taxonomy.py
```diff
@@ -54,17 +55,22 @@ class Taxonomy:
...
     with open(taxonomy_file_path, encoding="utf-8") as f:
         raw = yaml.safe_load(f)
     if "taxonomy" not in raw:
         raise ValueError(f"taxonomy.yml file {taxonomy_file_path} is missing the 'taxonomy' key at the root")
+    logger.info(f"Taxonomy YAML parsed successfully from {taxonomy_file_path}")
```
tbittar: debug

gems_views_builder/views.py
```diff
@@ -57,63 +58,93 @@ def create_intermediate_dir(self) -> Path:
...
+                        parquet_files_to_process.append(temp_metric_view)
+
+                        logger.info(f"[{metric_id}] Cleaning intermediate files")
+                        self._clean_intermediate_metric(metric_view_parquet_path, metric_structure_path)
+
+            logger.info("Cleaning filtered simulation table")
+            self._clean_filtered_simulation_table(filtered_simulation_table_path)
+
+            with logger.use_context("writer"):
+                logger.info("Step 3: Consolidating results")
```
tbittar: Not very helpful, what about (probably the consolidate_results function should be renamed)
```suggestion
                logger.info("Step 3: Writing results")
```

tbittar:
## Code Review — `feat(logging): add PipelineLogger with context management`

### Overview

Adds structured logging across all modules: every load, parse, and aggregation step emits `INFO` logs. The centrepiece is `PipelineLogger` in `common.py`, which wraps `logging.Logger` to add per-handler `ContextFilter` labels and a file-handler lifecycle (`configure_pipeline_run` / `close_pipeline_run`).

---

### The core question: is `PipelineLogger` overkill?

**Yes — it can be collapsed to ~20 lines.** Here's what it provides vs. what's needed:

| Feature | Needed? | Simpler alternative |
|---|---|---|
| Timestamped file handler per run | Yes | Module-level `configure_pipeline_run()` function |
| `close_pipeline_run()` on finish/failure | Yes | Stays as a module-level function |
| `[context]` label switchable via `use_context()` | Marginal | The messages already embed context (`[metric_id]`, step numbers) |
| Proxy methods (`info`, `warning`, `error`) | No | Call `logger.info()` directly on a plain `Logger` |
| `_has_named_handler` duplicate-handler guard | Yes (needed for tests) | One `if not logger.handlers:` check |

The `use_context()` context manager and `ContextFilter` machinery are the most complex parts, and they're used *inconsistently* — most log calls in `views.py` sit outside any `use_context()` block and will just emit `[gems views builder]`. The context info is already in the message text.

**Proposed replacement** (drops `PipelineLogger` entirely):

```python
# common.py
import logging, sys
from datetime import datetime, timezone
from pathlib import Path
from typing import Literal

PARQUET_COMPRESSION: Literal["zstd"] = "zstd"
PARQUET_COMPRESSION_LEVEL = 3
PARQUET_ROW_GROUP_SIZE = 64_000

_LOG_FORMAT = "[%(asctime)s][%(levelname)s] %(message)s"

logger = logging.getLogger("gems_views_builder")
logger.setLevel(logging.INFO)
logger.propagate = False
if not logger.handlers:
    _console_handler = logging.StreamHandler(sys.stdout)
    _console_handler.setFormatter(logging.Formatter(_LOG_FORMAT, datefmt="%Y-%m-%d %H:%M:%S"))
    logger.addHandler(_console_handler)

_file_handler: logging.FileHandler | None = None

def configure_pipeline_run(input_data_path: Path) -> Path:
    global _file_handler
    close_pipeline_run()
    logs_dir = input_data_path / "logs"
    logs_dir.mkdir(parents=True, exist_ok=True)
    timestamp = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%S%f")
    log_path = logs_dir / f"pipeline-{timestamp}.log"
    _file_handler = logging.FileHandler(log_path, encoding="utf-8")
    _file_handler.setFormatter(logging.Formatter(_LOG_FORMAT, datefmt="%Y-%m-%d %H:%M:%S"))
    logger.addHandler(_file_handler)
    return log_path

def close_pipeline_run() -> None:
    global _file_handler
    if _file_handler is None:
        return
    logger.removeHandler(_file_handler)
    _file_handler.close()
    _file_handler = None
```

All call sites stay the same (`logger.info(...)`, `configure_pipeline_run(...)`, `close_pipeline_run()`). Remove `logger.use_context(...)` blocks in `views.py` — the inner log messages already carry the context.

---

### Other issues

**Log verbosity / level misuse** — several `INFO` lines should be `DEBUG` or removed:
- `"Parsing catalog YAML from {path}"` + `"Catalog YAML parsed successfully from {path}"` bracket a single `yaml.safe_load` call; one line is enough and it's low-value `DEBUG`
- Same before/after pattern in `taxonomy.py`, `library.py`, `metrics.py`
- `get_catalog_metric` logs "Looking up X" *then* "X found" — verbose for a dict lookup

**Double-logging in `views.py`** — after this PR, loading a catalog emits ~6 lines (2 from `_load_catalog_file`, 2 from `load_catalog`, 1 from `get_catalog_metric`, 1 from `ViewConfig.load_catalog`). Inner functions shouldn't log when the outer orchestrator already does.

**`use_context()` scope bug** in `views.py`:
```python
with logger.use_context(f"catalog:{catalog_id}"):
    logger.info(f"Processing catalog '{catalog_id}'")
# context is reset here — but the actual catalog/metric work continues outside the block
catalog = self.loader.view_config.load_catalog(catalog_id)
```
The context switch exits before the actual work, so it has no effect on any of the inner log calls.

---

### Verdict

- **Remove `PipelineLogger`** — replace with the ~20-line module-level approach above.
- **Remove `use_context()` / `ContextFilter`** — not providing real value given inconsistent use.
- **Demote noisy lines to `DEBUG`** or remove the redundant before/after pairs.

The instrumentation breadth is good; the issue is the machinery, not the intent.

gems_views_builder/loader.py
dusanparipovic: I found that I have double logs of everything e.g. inside method and outside, so I removed all logs inside outside loading methods everything is inside methods.

dusanparipovic (dismissed):

aoustry (dismissed):

pyproject.toml
guilpier-code, line 39:
```txt
"pytest>=7.0.1",
```
Trying to run tests in **test_filter_simulation_table.py**, I faced a crash.
My python version is **3.14**, and **pytest 7.0.1** seems too old.
Digging a bit : it seems that for **python > 3.8**, **pytest** needs to be **> 8.0**, otherwise a crash occurs.
I had to : `pip install --upgrade pytest` to make it work.

gems_views_builder/views.py
guilpier-code, line 34: **Main program** : 
Looking for occurrences of **ViewBuilder**, I expect to reach the main program.
It seems there is no.
Is this normal ?

This **main program** should take the path to input data as argument and possibly set the logging configuration with this path, among other things.

gems_views_builder/views.py
```diff
@@ -18,6 +18,7 @@
 
 from gems_views_builder.aggregator import Aggregator
```
guilpier-code, line 20: _**Cosmetic**_ 
This source file contains mainly a class called **ViewBuilder**, but is called **views.py**.
Why not call it **views_builder.py** ?

gems_views_builder/views.py
guilpier-code, line 36: **Design**

Here we are in the **heart of the whole algorithm**.
If there is a place for high code quality, it's here.
Code is already quite clear and highlights correlations with the specifications.

But responsibilities should be also well-separated, and every software entity (class or function) should strictly stick to its mission.
The benefit of this is (at least) that : 
- code entities get smaller and easier to test in isolation
- if code changes in a part of the code, the rest of it remains unchanged, so dependencies between software entities are reduced
- code is more changeable 

According to this, **ViewBuilder**  makes too many things, for example :
- class **ViewBuilder** reads its input data, whereas it should  receive it already loaded and checked.
- class **ViewBuilder** writes resulting view whereas it should just return a this view.
- It also configures and closes the pipeline logs whereas these tacks should be done outside, and before creating a view builder.

For more details on these points, see others remarks in this file.

guilpier-code, line 50: "Private" methods **_clean_intermediate_metric** and **_clean_filtered_simulation_table** don't use any data member. So they should be free helper functions.

guilpier-code:
Do we have to embed file **uv.lock** in the git repo ? Or is it a generated file we don't want ?

gems_views_builder/views.py
```diff
@@ -57,63 +58,88 @@ def create_intermediate_dir(self) -> Path:
...
-                )
-
-                parquet_files_to_process.append(temp_metric_view)
-
-                self._clean_intermediate_metric(metric_view_parquet_path, metric_structure_path)
-
-        self._clean_filtered_simulation_table(filtered_simulation_table_path)
-
-        self.writer.consolidate_results(parquet_files_to_process)
+        pipeline_log_path = configure_pipeline_run(self.input_data_path)
```
guilpier-code: **logging** config should not be done here.
About this, please see [review of gems_views_builder/common.py](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/49#discussion_r3394544750)

gems_views_builder/common.py
```diff
@@ -1,5 +1,44 @@
...
 from typing import Literal
 
 PARQUET_COMPRESSION: Literal["zstd"] = "zstd"
 PARQUET_COMPRESSION_LEVEL = 3
 PARQUET_ROW_GROUP_SIZE = 64_000
+
+_LOG_FORMAT = "[%(asctime)s][%(levelname)s] %(message)s"
+_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+logger = logging.getLogger("gems_views_builder")
```
guilpier-code: This logging configuration seems complicated, and maybe we need all this.
So I have several remarks or questions : 
- Builtin module **logging** is a singleton, so it allows to set its configuration anywhere we want. This configuration is often set in the main program. In tests about a software entity involving logs (ex : **ViewBuilder**), we may even set **logging** so that it does nothing (if we want) : 
```python
# Disable all logging calls of severity CRITICAL and below
logging.disable(logging.CRITICAL)
```
Anyway, this logging config setting should not be done inside a software entity like **ViewBuilder**, it's not the mission / responsibility of **ViewBuilder**. Its mission is strictly to build views.
- Setting **logging** configuration currently seems to imply adding **timestamps** and give file to log in, so why not have something like: 
```python
import logging
...
# Assuming that log_dir is a Path
log_file = log_dir / 'application.log'

logging.basicConfig(
    level=logging.INFO,
    format='%(asctime)s - %(levelname)s - %(message)s',
    filename=str(log_file),  # Convert Path to string
    filemode='a'
)

# Usage
logging.info("This message includes a timestamp")
```
?
It seems more simple like this.

tbittar:
Yes it is meant for it, so that each user cloning the repo can replicate the excact same environment locally : see https://docs.astral.sh/uv/guides/projects/#uvlock

gems_views_builder/views.py
guilpier-code, line 41: According to what was said earlier, we should not have **input_data_path** passed to constructor, but rather **input_data** itself.
**input_data** could be a class that holds : 
- input_data.simulation_table
- input_data.view_config
- input_data.taxonomy
- input_data.calendar
- ...

We could also pass directly to constructor all these elements one by one (simulation_table, view_config, taxonomy, ...).

gems_views_builder/views.py
```diff
@@ -57,63 +58,88 @@ def create_intermediate_dir(self) -> Path:
...
+        try:
+            logger.info(f"Pipeline log file initialized at {pipeline_log_path}")
+            logger.info(f"Starting pipeline for study at {self.input_data_path}")
```
guilpier-code: Logging on pipeline logging itself is useless : path to log file is supposed to be given in documentation.
We should remove these lines.

gems_views_builder/views.py
```diff
@@ -57,63 +58,88 @@ def create_intermediate_dir(self) -> Path:
...
+
+            intermediates_dir = self.create_intermediate_dir()
+            filtered_simulation_table_path = intermediates_dir / "simulation_table_filtered.parquet"
+
+            # # 1. Filter simulation table (written to disk)
+            logger.info("Step 1: Filtering simulation table by calendar")
+            self.loader.simulation_table.filter_simulation_table(
```
guilpier-code: About : 
```python
self.loader.simulation_table.filter_simulation_table(...)
```
I have several problems with this method : 
- this method is (indirectly) part of a loader. But this loader (via this method) also writes on disk. It loads and it also writes. There is a responsibility problem here.
- This method returns something (a **FilteredSimulationTable** object) that is never caught / used. So why return something ?
- When looking for **FilteredSimulationTable** in base code, it seems to be used nowhere. I may make a mistake, do I ?
- As already said, we write on disk and then read the written files, which is not unnatural and slow. There must be better to do.

What I recommend : 
- For first remark, as already said above, the concept of loading should be banned from class **ViewBuilder** and we should have already loaded data as input of class **ViewBuilder**.
- Of course avoid reading from / writing to disc whenever possible.
- Remove class **FilteredSimulationTable** (and related tests) if it turns out we don't need it.

gems_views_builder/views.py
```diff
@@ -57,63 +58,88 @@ def create_intermediate_dir(self) -> Path:
...
+                for metric_id in metrics:
+                    logger.info(f"[{metric_id}] Processing metric")
+                    try:
+                        metric: Metric = get_catalog_metric(catalog, metric_id)
+                    except ValueError:
+                        logger.info(f"[{metric_id}] Metric not found in catalog '{catalog_id}' — skipping")
+                        continue  # # We should decide do we want to break process fully or continue without the current metric
+
+                    # # 2.3 Build metric structure table, persist to disk, then re-open lazily
+                    metric_structure_table = MetricStructureBuilder(
```
guilpier-code: **Caution** : _this remark was done without knowing that dumping some data on disk for saving RAM reasons  is **Rte** requirements_

Same remark as above, but for the **metric structure table** : we should avoid writing to / reading from disc (whenever possible). 

We should ban ban any writer or loader in class **ViewBuilder**.

gems_views_builder/catalog.py
```diff
@@ -110,24 +111,33 @@ def _to_metric(metric_data: MetricData) -> Metric:
 
 
 def load_catalog(catalog_file_path: Path) -> Catalog:
+    logger.info(f"Loading catalog from {catalog_file_path}")
```
guilpier-code: many `logger.info(....)` here. 
Do we need all this logging ?
Or should we rather log for bigger steps and only logs for details in case of exception raised or warnings ?

gems_views_builder/views.py
```diff
@@ -57,63 +58,93 @@ def create_intermediate_dir(self) -> Path:
...
+                with logger.use_context(f"catalog:{catalog_id}"):
+                    logger.info(f"Processing catalog '{catalog_id}' ({len(metrics)} metric(s))")
+                # # 2.1 Load catalog
+                catalog: Catalog = self.loader.view_config.load_catalog(catalog_id)
+                # # 2.2 Iterate over all metrics for this catalog
+                for metric_id in metrics:
+                    with logger.use_context(f"metric:{metric_id}"):
+                        logger.info(f"[{metric_id}] Processing metric")
+                        try:
+                            metric: Metric = get_catalog_metric(catalog, metric_id)
```
guilpier-code: About : 
```python
metric: Metric = get_catalog_metric(catalog, metric_id)
```
**get_catalog_metric** is a free function but it would make sense to make it a method of class **Catalog**.
We could have instead : 
```python
metric = catalog.get_metric(metric_id)
```

gems_views_builder/views.py
```diff
@@ -57,63 +58,93 @@ def create_intermediate_dir(self) -> Path:
...
+                        except ValueError:
+                            logger.info(f"[{metric_id}] Metric not found in catalog '{catalog_id}' — skipping")
+                            continue  # # We should decide do we want to break process fully or continue without the current metric
+
+                        # # 2.3 Build metric structure table, persist to disk, then re-open lazily
+                        metric_structure_table = MetricStructureBuilder(
+                            self.loader.system,
+                            catalog,
```
guilpier-code: we don't need **catalog** here : class **MetricStructureBuilder** don't use it.

gems_views_builder/views.py
```diff
@@ -57,63 +58,93 @@ def create_intermediate_dir(self) -> Path:
...
+
+                        logger.info(f"[{metric_id}] Cleaning intermediate files")
+                        self._clean_intermediate_metric(metric_view_parquet_path, metric_structure_path)
+
+            logger.info("Cleaning filtered simulation table")
+            self._clean_filtered_simulation_table(filtered_simulation_table_path)
+
+            with logger.use_context("writer"):
+                logger.info("Step 3: Consolidating results")
+                self.writer.consolidate_results(parquet_files_to_process)
```
guilpier-code: As said before, writing the final view is not the mission of the **ViewBuilder**. The **ViewBuilder** only builds a view.
The **build(...)** method should return the **final view**, the one that will be printed with a writer, out of this class, for example in the main program.

gems_views_builder/views.py
```diff
@@ -57,63 +58,93 @@ def create_intermediate_dir(self) -> Path:
...
+                for metric_id in metrics:
+                    with logger.use_context(f"metric:{metric_id}"):
+                        logger.info(f"[{metric_id}] Processing metric")
+                        try:
+                            metric: Metric = get_catalog_metric(catalog, metric_id)
+                        except ValueError:
+                            logger.info(f"[{metric_id}] Metric not found in catalog '{catalog_id}' — skipping")
+                            continue  # # We should decide do we want to break process fully or continue without the current metric
+
+                        # # 2.3 Build metric structure table, persist to disk, then re-open lazily
```
guilpier-code: Building the **metric structure table** is part **2.1** of the specifications, not **2.3** (in the version I have)

gems_views_builder/catalog.py
```diff
@@ -110,24 +111,33 @@ def _to_metric(metric_data: MetricData) -> Metric:
...
+    logger.info(f"Parsing catalog YAML from {catalog_file_path}")
     with open(catalog_file_path, encoding="utf-8") as f:
         raw = yaml.safe_load(f)
     if "catalog" not in raw:
         raise ValueError(f"catalog.yml file {catalog_file_path} is missing the 'catalog' key at the root")
+    logger.info(f"Catalog YAML parsed successfully from {catalog_file_path}")
     return CatalogData.model_validate(raw["catalog"])
 
 
 def get_catalog_metric(catalog: Catalog, metric_id: str) -> Metric:
```
guilpier-code: **get_catalog_metric** : should be a method **get_metric** of class **Catalog** (already said in **views.py**'s review)

gems_views_builder/catalog.py
```diff
@@ -110,24 +111,33 @@ def _to_metric(metric_data: MetricData) -> Metric:
 
 
 def load_catalog(catalog_file_path: Path) -> Catalog:
```
guilpier-code, line 128: **load_catalog** : this is the catalog loader, this is good : must remain out of class **Catalog**.

gems_views_builder/library.py
guilpier-code, line 64: Still in the idea of separating responsibility, I would have separated **ModelLibrary** and loading model library.
So **ModelLibrary** would not have loading methods, and the loader would table a path and return a loaded model library.

gems_views_builder/loader.py
guilpier-code, line 12: Still in the idea of separating responsibility, I would define a class **InputData** and a class **Loader**.
The `Load.load(input_data_path)` would return a **InputData**.
**InputData** would be a container for **system**, **taxonomy**, **view_config**, ....
An object of last class would be passed to **ViewBuilder**'s constructor (in the main program)

gems_views_builder/metrics.py
guilpier-code, line 61: Same remark as before : we should separate class **ViewConfig** from its loader.
```python
input_path = Path(...)
...
view_config_loader = ViewConfigLoader()
view_config = view_config_loader.laod(input_path)
...
```

guilpier-code, line 61: Why don't class **ViewConfig** have its own source file  **view_config.py**

gems_views_builder/views.py
```diff
@@ -57,63 +58,93 @@ def create_intermediate_dir(self) -> Path:
...
+                            self.loader.taxonomy,
+                            self.loader.model_library,
+                        ).build()
+
+                        metric_structure_path = self.writer.write_metric_structure_table(
+                            metric_structure_table.dataframe, metric.id
+                        )
+
+                        filtered_simulation_table_lazy = pl.scan_parquet(filtered_simulation_table_path)
+                        metric_structure_lazy = pl.scan_parquet(metric_structure_path)
```
guilpier-code: Creating **filtered_simulation_table_lazy** could be made once for all outside (and before) the nested loops over **catalog** and **metrics**. Otherwise we create it several times.

gems_views_builder/views.py
```diff
@@ -57,63 +58,93 @@ def create_intermediate_dir(self) -> Path:
         return intermediates_dir
 
     def build(self) -> None:
```
guilpier-code, line 62: The **build()** method is the heart of out work here, it's important that it's clear, as short and simple as possible, modular and testable in isolation the more we can.
Thinking about it a bit harder, I suggest some ideas for this.
This idea is to highlight the core of the algorithm and reject details in classes or functions. 

Here are details we don't want to know about at a high level (in `ViewBuilder.build()`) :
- loading input data
- intermediate data persistence
- persistence format (parquet)
- storage when data is retrieved (polars dataFrame), if possible
- even logging should not be done at this level, but internally to classes and functions, whenever possible 

You're free to accept these changes or not.
If you do, this will require a bit of work but code will get better. New ideas to make code cleaner and more modular  are welcome of course : you are at the best place to do it, because you know the code and specifications better.

What we can do : 
- As said before, remove logging configuration from `build()`
- About **filtered simu table** : you know what ? We should build a **filtered simu table** before creating the **ViewBuilder**, and pass it to the **ViewBuilder**'s constructor. The class **FilteredSimuTable** (which currently exists but is unused) will hide all the details it contains : any client code  of this class (like `ViewBuilder.build()`) won't know that this table is written and persistent somewhere on disc (**FilteredSimuTable** constructor will take **input_path** as argument and is the only one who know where it persists on disc), or ideally that it's retrieved as a polars **dataframe** (not sure it's possible, so a public method `dataFrame()` seems mandatory). 
  By doing this, notion of **intermediates_dir** or parquet format disappears from class **ViewBuilder**. Notion of **calendar** also disappears from `build()`. 
- We could do the same with the metric structure class : when manipulating an object of class **MetricStructure**, the client code should not know what's happening behind the scene (persistence, storage format, ...)
- As already said, loading input data should not be done in `ViewBuilder.build()` but input data should be passed already loaded and checked.
- the output of `build()` could be a **MergedView**.

In the end, `build()` would be an overview of the process of building the view, with no details we don't want to know about. We reject details in classes or methods.
We could have (something like) : 
```python
MergedView view().
for catalog_id, metrics in ...
    ...
    for metric_id in metrics:
        metric = catalog.get_metric(metric_id)
        metric_structure_table = MetricStructureBuilder(self.system, metric, self.taxonomy, self.model_library).build()
        metric_view = join_by_group(metric_structure_table, self.filtered_simu_table)
        metric_view.termsAggregation(metric)
        metric_view.temporalAggregation(metric)
        view.add(metric_view)
     view.merge()
     return view
```

pyproject.toml
tbittar: @guilpier-code can you open an issue with this ? We should also drop support of Python 3.10 for which support ends in october : https://devguide.python.org/versions/

tbittar: I guess the `uv.lock` should pin a Python version with `uv python pin` ?

gems_views_builder/views.py
tbittar: Writing on disk is done for the simulation table to split it in several chuncks, but during the view building process, I do not see how writing the intermediate file to disk could lead to RAM improvements (since we read it afterwards). The only thing that cold save RAM is to have a mechanism to build the views by chuncks (by it is not obvious how to do this)

pyproject.toml
guilpier-code: The corresponding **gitHub** issue can be found [here](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/63)

gems_views_builder/library.py
dusanparipovic: Done #66

gems_views_builder/loader.py
dusanparipovic: Done #66

gems_views_builder/metrics.py
dusanparipovic: Done #66

dusanparipovic: Done #66

gems_views_builder/views.py
dusanparipovic: Done #66

gems_views_builder/views.py
dusanparipovic: Done #66

dusanparipovic: Done! #66

dusanparipovic: Resolved #66

gems_views_builder/views.py
dusanparipovic: Done! #66

gems_views_builder/views.py
guilpier-code: no more relevant

gems_views_builder/validation/catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,45 @@
...
+        validate_catalog_against_taxonomy(catalog, taxonomy)
+    logging.info(f"All catalogs are consistent with taxonomy {taxonomy.id!r}")
+
+
+def _category_ports_by_id(taxonomy: Taxonomy) -> dict[str, set[str]]:
+    return {category.id: {port.id for port in category.ports} for category in taxonomy.categories}
+
+
+def validate_catalog_against_taxonomy(catalog: Catalog, taxonomy: Taxonomy) -> None:
+    logging.info(f"Validating catalog {catalog.id!r} against taxonomy {taxonomy.id!r}")
```
tbittar, line 21: Could be debug level

gems_views_builder/validation/study_layout_validator.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class StudyLayoutValidator:
+    """
+    Expected files:
+    - taxonomy.yml
+    - view_config.yml
+    - library.yml
+    - system.yml
+    - simulation_table.parquet
```
tbittar: Should possibly be csv

tests/test_catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,78 @@
...
+    validate_catalog_against_taxonomy(catalog, taxonomy)
+
+
+def test_validate_catalogs_against_taxonomy_passes_for_test_dataset(test_dataset_dir: Path) -> None:
+    from gems_views_builder.metrics import ViewConfig
+
+    taxonomy = load_taxonomy(test_dataset_dir / "taxonomy.yml")
+    view_config = ViewConfig.load(test_dataset_dir / "view_config.yml")
+    catalogs = load_catalogs(test_dataset_dir, view_config.catalog_ids)
+    validate_catalogs_against_taxonomy(catalogs, taxonomy)
```
tbittar, line 37: These 2 tests seems a bit redundant

dusanparipovic: Yes right I agree in fact we have only one catalog, inside test directory. 
I'll add dedicated tests for multiple catalogs inside `catalog validation branch`

aoustry (approved):

guilpier-code (approved):

PR #50, opened 2026-06-04
feat(validation): add catalog<->taxonomy validation

tbittar:
## Summary

- Add `validation/` package with two validators:
  - `StudyLayoutValidator` — checks all required files exist before loading begins; replaces the old `InputValidator`
  - `validate_catalogs_against_taxonomy` — cross-checks every catalog against the study taxonomy (matching id, category names, and port names)
- `Loader` eagerly loads all catalogs referenced in `view_config` into `self.catalogs: dict[str, Catalog]`, validated once and reused throughout the pipeline
- `ViewBuilder.__init__` wires both validators in sequence; `build()` reads catalogs from `loader.catalogs` instead of calling `view_config.load_catalog` per loop iteration

## Test plan

- [x] `test_catalog_taxonomy_validator.py` — all new validator unit tests pass
- [x] `test_views_input_checks.py` — input layout validation tests pass with new `StudyLayoutValidator`
- [x] End-to-end `test_view_builder.py` still passes (no regression)


tbittar (commented):

**Dead code** (`metrics.py` — `ViewConfig.load_catalog`): This method is no longer called since `Loader` now eagerly loads all catalogs via `load_catalogs()` into `self.catalogs`. It should be removed to avoid confusing future readers who see two catalog-loading paths.

gems_views_builder/loader.py
```diff
@@ -33,9 +35,8 @@ def load_into_self(self) -> "Loader":
             self.input_data_path / "library.yml"
         )  # # must be named like this for now, in future when we enable user to have more than one libraries we should decide pattern to use
         logger.info("Model library loaded")
+        system_path = next(self.input_data_path.glob("system*"))
```
tbittar: **Unsafe glob**: `next(input_data_path.glob("system*"))` will match any file whose name starts with `system` — e.g. `system_backup.yml` or `system_old.yml`. If such a file is present, the wrong system is silently loaded. Consider checking explicitly for `system.yml` first, or asserting that exactly one match exists and raising a descriptive error when there are none or more than one.

dusanparipovic: Agree this is fixed, I'll write issue which will propose new folder structure and that will solve problem of file names/finding.

gems_views_builder/validation/catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,44 @@
...
+    category_ports_by_id = _category_ports_by_id(taxonomy)
+
+    for metric in catalog.metrics.values():
+        for term in metric.terms:
+            if term.taxonomy_category not in category_ids:
+                raise ValueError(
+                    f"Catalog {catalog.id!r} metric {metric.id!r} uses taxonomy-category "
+                    f"{term.taxonomy_category!r}, which is not defined in taxonomy {taxonomy.id!r}"
+                )
+
```
aoustry, line 36: @dusanparipovic: and what about the output-id of the term? We would check whether it is a variable, a port field or an extra-output of the taxonomy.

tests/test_catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def test_validate_catalog_against_taxonomy_raises_on_unknown_location_port(test_dataset_dir: Path) -> None:
+    taxonomy = load_taxonomy(test_dataset_dir / "taxonomy.yml")
+    catalog = load_catalog(next((test_dataset_dir / "catalogs").glob("*.yml")))
+    next(iter(catalog.metrics.values())).terms[0].location_ports = "unknown_port"
+    with pytest.raises(ValueError, match="uses location-port"):
+        validate_catalog_against_taxonomy(catalog, taxonomy)
+
+
```
aoustry, line 64: Add also tests for taxonomy-categories and output-id inconsistencies (see comments above)

gems_views_builder/validation/study_layout_validator.py
```diff
@@ -0,0 +1,79 @@
+from pathlib import Path
+
+from gems_views_builder.common import logger
+
+EXACT_FILES = ["taxonomy.yml", "view_config.yml", "library.yml", "system.yml"]
+PREFIX_FILES = {"calendar": ".csv", "simulation_table": ".parquet"}
+
+
+class StudyLayoutValidator:
```
aoustry, line 9: @tbittar do you think we should keep this test file, as we will authorize more flexibility regarding the ViewBuilder inputs in the near future?

gems_views_builder/validation/catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,44 @@
...
+    return {category.id: {port.id for port in category.ports} for category in taxonomy.categories}
+
+
+def validate_catalog_against_taxonomy(catalog: Catalog, taxonomy: Taxonomy) -> None:
+    logger.info(f"Validating catalog {catalog.id!r} against taxonomy {taxonomy.id!r}")
+    if catalog.taxonomy != taxonomy.id:
+        raise ValueError(
+            f"Catalog {catalog.id!r} references taxonomy {catalog.taxonomy!r}, but study taxonomy id is {taxonomy.id!r}"
+        )
+
```
aoustry, line 25: @dusanparipovic: and what about the taxonomy-categories mentioned in the catalog? They should be checked as well.

aoustry (changes requested):

guilpier-code:
Do we need to onboard file **uv.lock** in our git repo ?

guilpier-code:
We also onboard file **resources/tests/test_inputs/test_3/simulation_table--20260318-0623.parquet**
It could have a more generic name, like **simulation_table--yyyymmdd-hhss.parquet**

resources/tests/test_inputs/test_3/catalogs/catalog.yml
```diff
@@ -0,0 +1,60 @@
...
+  - id: LOAD
+    terms:
+      - taxonomy-category: consumption
+        output-id: active_load
+        location-ports: p_balance_port
+    terms-operator: sum
+    time-operator: sum
+
+  
+  - id: INSTALLED_PROD_CAPACITY
```
guilpier-code, line 40: **Simplicity**

The metric **INSTALLED_PROD_CAPACITY** is unused in this test. 
It could be removed from this file.

resources/tests/test_inputs/test_3/library.yml
```diff
@@ -0,0 +1,527 @@
...
+
+# Library of models for the emulation of PyPSA
+# References for PyPSA
+# T. Brown, J. Hörsch, D. Schlachtberger, PyPSA: Python for Power System Analysis, 2018, Journal of Open Research Software, 6(1), arXiv:1707.09913, DOI:10.5334/jors.188
+# https://pypsa.readthedocs.io/en/latest/
+
+library:
+  id: pypsa_models
+  description: PyPSA model library 
+  #https://pypsa.readthedocs.io/en/latest/user-guide/components.html#
```
guilpier-code, line 21: Commented **urls** lead to **404 not found**. 
We should correct them or remove them (the latter may be a better solution)

resources/tests/test_inputs/test_3/library.yml
```diff
@@ -0,0 +1,527 @@
...
+        - id: p0_lower
+          expression: p0 >= p_min_pu * p_nom
+      objective-contributions:
+        - id: capital_objective
+          expression: capital_cost * p_nom
+        - id: operational_objective
+          expression: sum(marginal_cost * p0) 
+
+  
+    - id: storage_unit
```
guilpier-code, line 296: model **storage_unit** is unused, we could remove it

resources/tests/test_inputs/test_3/library.yml
```diff
@@ -0,0 +1,527 @@
...
+        - id: p_max_pu
+          time-dependent: true
+          scenario-dependent: true
+        - id: capital_cost 
+          time-dependent: false
+          scenario-dependent: false
+        - id: marginal_cost
+          time-dependent: true
+          scenario-dependent: true
+        #- id: marginal_cost_quadratic #Parameter not instantiated for now
```
guilpier-code, line 239: We could remove all the commented lines, useless

resources/tests/test_inputs/test_3/library.yml
```diff
@@ -0,0 +1,527 @@
...
+          expression: state_of_charge <= max_hours * p_nom
+        - id: state_of_charge_balance
+          expression: state_of_charge = (1- standing_loss) * state_of_charge[t-1] + efficiency_store * p_store - p_dispatch / efficiency_dispatch  + inflow - spill
+      objective-contributions:
+        - id: capital_objective
+          expression: capital_cost * p_nom
+        - id: operational_objective
+          expression: sum(marginal_cost * p_dispatch + spill_cost*spill + marginal_cost_storage * state_of_charge) 
+
+    - id: store
```
guilpier-code, line 406: model **store** useless

resources/tests/test_inputs/test_3/library.yml
```diff
@@ -0,0 +1,527 @@
...
+          expression: capital_cost * e_nom
+        - id: operational_objective
+          expression: sum(marginal_cost * p + marginal_cost_storage * e) 
+
+
+
+
+#### Models for PyPSA global constraint: https://pypsa.readthedocs.io/en/latest/user-guide/components.html#global-constraint ###
+
+    - id: global_constraint_co2_max
```
guilpier-code, line 493: model **global_constraint_co2_max** useless

gems_views_builder/simulation_table.py
guilpier-code, line 51: `class SimulationTable`
We should separate this class from its loading functions. 
There's a good reason for this : for now we only load simulation table in the **parquet** format.
But this format is unreadable for human, and, at least for test purpose, we should be able to read simulation table from a **csv** format : we would be able to make simulation tables by hand, and more easily control their content.

We could have 3 classes : **csvSTloader**, **parquetSTloader**, and **SimulationTable** (of course).
```python
cvs_ST_loader = csvSTloader()
simulation_table = cvs_ST_loader.load(path_to_csv_simu_table_file)

# or
 
parquet_ST_loader = parquetSTloader()
simulation_table = parquet_ST_loader .load(path_to_parquet_simu_table_file)
```
in both cases, class **SimulationTable** remains as small as possible, is agnostic to any loader, we could add a new loader without changing this class.

gems_views_builder/views.py
```diff
@@ -37,11 +38,11 @@ def __init__(
...
     ) -> None:
         self.input_data_path = input_data_path
-        # # 1. Validate input data, if it fails, the builder will not be able to build the views.
-        InputValidator(self.input_data_path).validate()
-        # # 2. Load input data, if it fails, the builder will not be able to build the views.
+        StudyLayoutValidator(self.input_data_path).validate()
         self.loader = Loader.load(self.input_data_path)
-        # # 3. Instantiate aggregator and writer if everything is loaded successfully
+        # # Currently we support only one taxonomy per study
+        validate_catalogs_against_taxonomy(self.loader.catalogs, self.loader.taxonomy)
```
guilpier-code, line 44: Again (see review of [PR 49](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/49)), validating the catalogs against taxonomy (validating input data in general) should not be done in class  **ViewBuilder**.
**ViewBuilder**'s input data should be loaded and checked before even creating a **ViewBuilder**, and this for responsibility separation reasons.

resources/tests/test_inputs/test_3/taxonomy.yml
```diff
@@ -0,0 +1,31 @@
+taxonomy:
+  id: my_taxonomy
```
guilpier-code, line 2: Taxonomy categories are useless : **coupling**, **storage**

tests/conftest.py
```diff
@@ -160,19 +70,16 @@ def test_dataset_dir(test_files_root: Path, request: pytest.FixtureRequest) -> P
 
 
 def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
```
guilpier-code, line 72: **pytest_generate_tests** and **_dataset_dirs** : are they used somewhere in tests ?
if not, we should consider removing them.

tests/test_catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,78 @@
...
+    validate_catalog_against_taxonomy,
+    validate_catalogs_against_taxonomy,
+)
+from gems_views_builder.views import ViewBuilder
+
+
+def test_validate_catalog_against_taxonomy_passes_for_test_dataset(test_dataset_dir: Path) -> None:
+    taxonomy = load_taxonomy(test_dataset_dir / "taxonomy.yml")
+    catalog = load_catalog(next((test_dataset_dir / "catalogs").glob("*.yml")))
+    validate_catalog_against_taxonomy(catalog, taxonomy)
```
guilpier-code, line 29: There is no **assert** part in this test. The **AAA** rule gives advice on unit tests structure (**Arrange** [building test material], **Act** [action to be tested], **Assert** [checks to be done after action]).
Here we expect no exception raised.
For more clarity, we could use : 
```python
import pytest
from contextlib import nullcontext as does_not_raise

def test_validate_catalog_against_taxonomy_passes_for_test_dataset(test_dataset_dir: Path) -> None:
    taxonomy = load_taxonomy(test_dataset_dir / "taxonomy.yml")
    catalog = load_catalog(next((test_dataset_dir / "catalogs").glob("*.yml")))
   
    # here Act and Assert a quite mixed, but it's ok
    with does_not_raise():
        validate_catalog_against_taxonomy(catalog, taxonomy)
```

tests/test_catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_validate_catalog_against_taxonomy_raises_on_unknown_location_port(test_dataset_dir: Path) -> None:
+    taxonomy = load_taxonomy(test_dataset_dir / "taxonomy.yml")
+    catalog = load_catalog(next((test_dataset_dir / "catalogs").glob("*.yml")))
+    next(iter(catalog.metrics.values())).terms[0].location_ports = "unknown_port"
+    with pytest.raises(ValueError, match="uses location-port"):
+        validate_catalog_against_taxonomy(catalog, taxonomy)
+
+
+def test_view_builder_raises_when_catalog_taxonomy_mismatch(
```
guilpier-code, line 65: This last test should not exist : indeed,  **ViewBuilder** should not load any data but it be passed already loaded data and checked input data (in the review of this [PR 49](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/49), I talk about this point in more details).

tests/test_catalog_taxonomy_validator.py
```diff
@@ -0,0 +1,78 @@
...
+def test_validate_catalog_against_taxonomy_passes_for_test_dataset(test_dataset_dir: Path) -> None:
+    taxonomy = load_taxonomy(test_dataset_dir / "taxonomy.yml")
+    catalog = load_catalog(next((test_dataset_dir / "catalogs").glob("*.yml")))
+    validate_catalog_against_taxonomy(catalog, taxonomy)
+
+
+def test_validate_catalogs_against_taxonomy_passes_for_test_dataset(test_dataset_dir: Path) -> None:
+    from gems_views_builder.metrics import ViewConfig
+
+    taxonomy = load_taxonomy(test_dataset_dir / "taxonomy.yml")
```
guilpier-code, line 35: In these tests, we should not load from files all materials we need to test functions : 
- `validate_catalog_against_taxonomy(catalog, taxonomy)` 
- or `validate_catalogs_against_taxonomy(catalogs, taxonomy)`

We should create objects from **Catalog**, **Taxonomy** or **ViewConfig** constructors.
It's : 
- more flexible : we don't need many **.yml** files to test many situations and cases. Here we are stuck with only one  **taxonomy.yml**, **view_config.yml** or **catalogs.yml**, which limits our options to build various tests
- simpler
- faster (no disc access).

tests/test_view_builder.py
```diff
@@ -10,12 +10,15 @@
...
+import logging
 import shutil
 from pathlib import Path
+from unittest.mock import patch
 
 import polars as pl
 import pytest
 
+import gems_views_builder.common as common_mod
 from gems_views_builder.views import ViewBuilder
```
guilpier-code, line 23: Here we're stuck with one integration study : **test_3**.
Integration tests / end-to-end tests like this one are necessary.
But what if we want to multiply testing situations, or test our code in more details ?
Will we be forced to have a set of **yaml** files each time we want to test a small behavior of the code ?
Of course not.

That's the reason why [this suggestions](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/49#discussion_r3403515482) were made. 
It's a trial to make **ViewBuilder** more modular.
The idea is to break **ViewBuilder** in several pieces, in order to be able to test each component behaviors in isolation.

Testing **ViewBuilder** is also testing each piece of code it's made of.
It allows smaller tests, we can target more precisely what is to be tested.
So testing **ViewBuilder** is having a lot of short and quick tests.

I'm not saying that integration tests are not required, they are. But they must be far less numerous than unit tests.

gems_views_builder/metrics.py
guilpier-code, line 66: `Apart from the fact that ViewConfig and the related Loader should be 2 separated (a class and a function for instance)`

Constructor of **ViewConfig** and **load** method both take **config_file_path** as argument.
We should choose. My choice goes to **load** method : class **ViewConfig** should not know from which source data it was built, only the load function should / loader should know.
We should remove data member `self.file`

resources/tests/test_inputs/test_3/library.yml
tbittar: Librairies are a collection of models that we often dulicated between repos. Probably for these tests some libs are not used, but it could be in the future in we enrich the test case

gems_views_builder/simulation_table.py
tbittar: i agree but why not a single function that detects the file extension and calls the correct loading method ? Otherwise have a interface from which both loader class derive

guilpier-code: Yes, both solutions can be implemented. If it turns out both loading codes have more than a path in common, inheritance can be chosen.

PR #51, opened 2026-06-04
feat(filter-breakdown): add metric filter/breakdown

### LINKED ISSUE #27: Filtering
enable the filtering in Metric definition, MetricStructure etc...

### LINKED ISSUE #28: Breakdown
Enable the breakdown with respect to properties in the Metric def, MetricStructure... etc

### LINKED ISSUE #58: fix: change metric location format
The current format for metric location use '{' and '}', which is misleading (this seems to relate to a set, whereas it is a n-uple, since the order is important. I propose the following design

- If metric_location is just unique string `"comp_id"` (case where location-port is a unique string): `"comp_id"`
- Else: `("comp_id1","comp_id2",...,"comp_idn)` if it is a n-uple, i.e. use **parentheses**.

tbittar:
## Summary

**GemsPy `properties` branch upgrade:**
- `library.py`: replace local Pydantic models with GemsPy `LibrarySchema`/`ModelSchema`; add `resolve_libraries()` for system resolution
- `system.py`: use `resolve_system()` + resolved `Component` type; `_get_peer_component` now returns `str | tuple[str, ...]` to support multiple peers per port; add `get_component()` helper; `from_file` renamed to `load(path, library)`
- `pyproject.toml`: pin `gemspy` to `GemsPy@properties` branch (required for `Component.properties`)

**Filter and breakdown:**
- `catalog.py`: add `PropertySchema(key, value?)`; replace `breakdown_property: str` and `filter: tuple` with `breakdown: list[PropertySchema]` and `filter: list[PropertySchema]`; Pydantic validator enforces at most one filter entry with a non-null value
- `metrics_builder.py`: `_check_filter_matches` skips components that do not match the metric filter; `_format_breakdown_properties` serialises selected property values into `breakdown_properties` as `{(key,value),...}`

## Test plan

- [ ] `test_filtering_and_breakdown.py` — regression tests for filter and breakdown outputs
- [ ] `test_filtering_and_breakdown_property_order.py` — ordering tests for multi-key breakdown strings
- [ ] `test_catalog.py`, `test_metric_structure_builder.py`, `test_input_system.py`, `test_model_library.py` — updated unit tests pass
- [ ] End-to-end `test_view_builder.py` with test_3 still passes


pyproject.toml
```diff
@@ -12,7 +12,8 @@ dependencies = [
     "pyarrow>=14.0.0",
     "pyyaml>=6.0.2",
     "pydantic>=2.11.7",
-    "gemspy>=0.0.4",
+    # GemsPy (PR #205 / `properties` branch): installs the `gems` top-level package.
+    "gemspy @ git+https://github.com/AntaresSimulatorTeam/GemsPy.git@properties",
```
tbittar: **Floating git ref**: Pinning to a branch name (`@properties`) makes the build non-reproducible — a force-push to that branch silently changes what gets installed. Replace with a commit SHA or a release tag as soon as `properties` is stabilised. Until then at least document the expected commit hash in the comment so CI failures are diagnosable.

gems_views_builder/input/catalog.py
```diff
@@ -44,13 +45,30 @@ class TermData(ViewBuilderBasedModel):
...
+
+
 class MetricData(ViewBuilderBasedModel):
     id: str
     terms: list[TermData]
     terms_operator: TermsOperator
     time_operator: TimeOperator
-    breakdown_property: str | None = None
-    filter: tuple[str, str] | None = None
+    breakdown: list[PropertySchema] | None = None
```
tbittar, line 60: **Misleading type for `filter`**: The field is typed `list[PropertySchema] | None` but the `validate_filter` validator immediately raises if `len(filter) > 1`. A single `PropertySchema | None` would express the intent directly without the surprising Pydantic validator. The current shape makes callers think multiple filter entries are valid.

gems_views_builder/metrics_builder.py
```diff
@@ -39,9 +40,28 @@ class MetricStructureTable:
...
 
-class MetricStructureBuilder:
-    """Build metric structure rows without loading unrelated datasets."""
+def _check_filter_matches(component: Component, filter: tuple[PropertySchema, ...] | None) -> bool:
+    if filter is None:
+        return True
+    return all(component.properties.get(entry.key) == entry.value for entry in filter)
+
+
+def _format_breakdown_properties(
```
tbittar: **Implicit serialisation contract**: The `{(key,value)}` format string is tested literally in many assertions across the test suite. If the format ever needs to change (e.g. to JSON or a different delimiter), it will be a wide blast radius. Consider a dedicated `BreakdownKey` type with a stable `__str__` or a module-level formatting constant so that the format is defined in exactly one place.

gems_views_builder/system.py
```diff
@@ -10,114 +10,184 @@
...
+
+            component_port_connections[(c1, p1)].add(c2)
+            component_port_connections[(c2, p2)].add(c1)
+
+        logger.info(f"Built component-port connection index with {len(component_port_connections)} entry(ies)")
         return component_port_connections
 
-    def _get_peer_component(self, component_id: str, port_id: str) -> str:
-        """Return the peer component id for (component_id, port_id), or raise ValueError."""
+    def _get_peer_component(self, component_id: str, port_id: str) -> str | tuple[str, ...]:
```
tbittar: **Misleading method name**: `_get_peer_component` (singular) now silently returns `str | tuple[str, ...]` when multiple peers exist. A caller reading only the method name would expect a single peer and would not anticipate handling a tuple. Rename to `_get_peer_components` or split into two methods to make the multi-peer case explicit.

tbittar (commented):

dusanparipovic:
Same as next [PR](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/52#issuecomment-4639815590), check comment

gems_views_builder/input/catalog.py
aoustry: @tbittar are you talking about the breakdown of the filter ? Because for breakdown, it makes sense to have a list. We already have such cases in the tests base.

aoustry: For `filter`, I agree that we can keep a single PropertySchema

gems_views_builder/input/catalog.py
```diff
@@ -78,8 +93,8 @@ class Metric:
     terms: list[Term]
     terms_operator: TermsOperator
     time_operator: TimeOperator
-    breakdown_property: str | None = None
-    filter: tuple[str, str] | None = None
+    breakdown: tuple[PropertySchema, ...] | None = None
```
aoustry: Why tuple ? I would say List...

tests/test_metric_structure_builder.py
```diff
@@ -111,17 +193,148 @@ def test_load_structure_component_and_location(
...
+    link_rows = df.filter(pl.col("component") == "link_link_AB")
+    assert len(link_rows) == 1
+    assert link_rows["metric_location"][0] == _format_metric_location(
+        system.get_location("link_link_AB", ("p0_port", "p1_port"))
+    )
+    assert set(_parse_metric_location(link_rows["metric_location"][0])) == {"busA", "busB"}
+    assert set(link_rows["output"].to_list()) == {"p0_port.flow"}
+
+
+def test_duplicate_locations_from_two_ports_produce_duplicate_rows(test_files_root: Path) -> None:
```
aoustry: The content of this tests looks correct, but I am not sure to understand the title of this test method ->this is not a duplication of rows that happens here, rather a tuple-ization of the metric_location id.

tests/test_metric_structure_builder.py
```diff
@@ -111,17 +193,148 @@ def test_load_structure_component_and_location(
...
+    bus_a_rows = df.filter(pl.col("component") == "busA")
+    assert len(bus_a_rows) == 1
+    assert bus_a_rows["metric_location"][0] == _format_metric_location(system.get_location("busA", "p_balance_port"))
+    assert set(_parse_metric_location(bus_a_rows["metric_location"][0])) == bus_a_expected
+
+    bus_b_expected = {"generator_B1", "link_link_AB"}
+    bus_b_rows = df.filter(pl.col("component") == "busB")
+    assert len(bus_b_rows) == 1
+    assert bus_b_rows["metric_location"][0] == _format_metric_location(system.get_location("busB", "p_balance_port"))
+    assert set(_parse_metric_location(bus_b_rows["metric_location"][0])) == bus_b_expected
```
aoustry: @dusanparipovic: I think this test makes no sense and should be replaced by another test.

Reminder: the catalog also defines a category for objects that should serve as location. This taxonomy category is unique. In this test, it looks like you have multiple "types" of components that serves as location...

aoustry: I think we should raise, in the code, an error 

- if there are multiple peers duing the computation of peers  _OR_ 
- if the peer is not in the allowed taxonomy category (location/taxonomy-category).

This corresponds to the comments I posted about the `get_location` function in system.py. I suggest that this method rather goes in metrics_builder.py, and that the method `_get_peer_components` remains in system.py.

aoustry: Hence you can replace this test by a test checking these two types of error.

gems_views_builder/input/system.py
```diff
@@ -10,114 +10,184 @@
...
+        # location_port is tuple[str, ...] — each named port resolves to one or more peers
         result: list[str] = []
         for port in location_port:
-            peer = self._get_peer_component(component_0_id, port)
-            result.append(peer)
+            peer = self._get_peer_components(component_0_id, port)
+            if isinstance(peer, str):
+                result.append(peer)
+            else:
+                result.extend(peer)
```
aoustry: You should raise en error if they are 0 peer of if they are multiple peers. Indeed, for each port in location_port, the locating peer component should be unique.

gems_views_builder/input/system.py
```diff
@@ -10,114 +10,184 @@
...
             return component_0_id
 
         if isinstance(location_port, str):
-            peer = self._get_peer_component(component_0_id, location_port)
+            peer = self._get_peer_components(component_0_id, location_port)
+            logger.debug(f"Resolved location for component {component_0_id!r} via port {location_port!r} to {peer!r}")
             return peer
 
-        # location_port is tuple[str, ...]
+        # location_port is tuple[str, ...] — each named port resolves to one or more peers
```
aoustry:
```suggestion
        # location_port is tuple[str, ...] — each named port resolves to a UNIQUE peer
```

gems_views_builder/input/system.py
```diff
@@ -10,114 +10,184 @@
...
         component_0_id: str,
         location_port: str | tuple[str, ...] | None,
     ) -> str | tuple[str, ...]:
         if location_port is None:
+            logger.debug(f"Using component {component_0_id!r} as location because no location port is defined")
             return component_0_id
 
         if isinstance(location_port, str):
-            peer = self._get_peer_component(component_0_id, location_port)
+            peer = self._get_peer_components(component_0_id, location_port)
```
aoustry: You should raise en error if they are 0 peer of if they are multiple peers. Indeed, for each port in location_port, the locating peer component should be unique.

gems_views_builder/input/system.py
aoustry: You should probably factorize the assertion checks of lines 151 and 158, for instance using recursivity: to manage the case where location_port is tuple, just sequentially apply `get_location` to each port individually (with the same argument `component_0_id`)

gems_views_builder/input/system.py
```diff
@@ -10,114 +10,184 @@
...
         result: list[str] = []
         for port in location_port:
-            peer = self._get_peer_component(component_0_id, port)
-            result.append(peer)
+            peer = self._get_peer_components(component_0_id, port)
+            if isinstance(peer, str):
+                result.append(peer)
+            else:
+                result.extend(peer)
+        logger.debug(
```
aoustry: This check should also assert that each unique peer (unique for each port) have the same taxonomy-category as the catalog.location.taxonomy-category (or a (grand)-child of it, in the near future).
For this reason, I suggest that get_location should rather be a method of the MetricBuilder.

_get_peer_components can remain in System.

gems_views_builder/input/system.py
aoustry: The `_get_peer_components` method does not raise an error itself because this is a general purpose method ->  in other applications, we may authorize multiple peers. So this is `get_location`, which is dedicated uniquely to metric building, that is responsible for raising such an error.

tests/test_input_system.py
```diff
@@ -13,104 +13,62 @@
...
-        system.get_location(cid, pid)
-    except ValueError as e:
-        if "Multiple connections found" in str(e):
-            pytest.xfail("Chosen ports have multiple peers in this dataset")
-        raise
-    pytest.fail("get_location should raise when (component, port) has multiple peers")
+    result = system.get_location(cid, pid)
+    expected_peers = system._component_port_connections[(cid, pid)]
+    assert isinstance(result, tuple)
+    assert set(result) == expected_peers
```
aoustry: This test should be changed: if multiple peer for a given location-port, the get_location method should raise an issue.

tests/test_input_system.py
```diff
@@ -13,104 +13,62 @@
...
```
aoustry, line 47: please also add a check to verify that get_location raises an error if 0 peer

tests/test_input_system.py
```diff
@@ -13,104 +13,62 @@
...
 
 
 def test_locating_function(test_dataset_dir: Path) -> None:
```
aoustry: The purpose of this test method is not clear. Clarify or delete.

aoustry (changes requested):

guilpier-code:
About new end-to-end test **resources/tests/test_inputs/filtering_and_breakdown** : 
Like suggested in [PR 50](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/50), don't hesitate to simplify this test material and remove useless things (models, taxonomy categories, ...)

gems_views_builder/input/catalog.py
tbittar: Yes and breakdown should keep a list

gems_views_builder/__init__.py
guilpier-code, line 56: From what I understand, defining `__all__` is useful when we import things of module **gems_views_builder** from another module, by doing : 
```py
from gems_views_builder import *
```
It narrows imported entities to elements in list `__all__`.
But I can't see any import like this in code base, leading to the idea that this `__all__` is useless.
Is it ?
Should we use it more often ?
Or maybe module **gems_views_builder** is meant to be used from another python code base ?
In that case, not using `__all__` in the current code base is expected.

guilpier-code, line 57: Same question as above for `__all__ `in **gems_views_builder/validation/__init__.py**

gems_views_builder/catalog.py
guilpier-code: `_to_metric` seems to be a **Metric** constructor. 
So we could expect it to be part of class **Metric**.

**Question** : 
Would it be possible to have : 
```py
@dataclass
class Metric:
    id: str
    terms: list[Term]
    terms_operator: TermsOperator
    time_operator: TimeOperator
    breakdown: tuple[PropertySchema, ...] | None = None
    filter: PropertySchema | None = None

    def __init__(self,metric_data: MetricData):
        self.id=metric_data.id,
        self.terms=[_to_term(term) for term in metric_data.terms],
        self.terms_operator=metric_data.terms_operator,
        self.time_operator=metric_data.time_operator,
        self.breakdown=tuple(metric_data.breakdown) if metric_data.breakdown else None,
        self.filter=metric_data.filter,
```
?

guilpier-code: Same question for `_to_term `as we ask for `_to_metric`

gems_views_builder/metrics_builder.py
guilpier-code: `self.catalog` not needed. 
Let's remove it.
Same for `self.taxonomy`, not required.

gems_views_builder/catalog.py
guilpier-code, line 104: **_Cosmetic_** 
**taxonomy_category** : the word **"taxon"** exists, means the same thing, but is quite shorter.
Should we replace **taxonomy category** by **taxon** in : **code**, **doc** and **input data** (*.yml files) ?

gems_views_builder/library.py
guilpier-code: **get_components_in_taxonomy_category(...)**
This method should be named **models_in_taxonomy_category(...)**.

gems_views_builder/metrics_builder.py
guilpier-code: This **build** function is a long and complex cognitively. It contains 3 nested loops and an internal if statement. 
We could try to shorten / simplify it by : 
- Remove one of the loop by : jumping directly from **term** to **component** (avoiding the loop on **models**). To do that, we could have a method / free function 
```python 
def components_in_taxon(model_library, system, term.taxonomy_category) -> list[str] : component_ids
```
- not checking input data : **_check_filter_matches** could be called in a input data validation phase (which currently exists if I remember correctly).
- There are 2 return statements, but maybe only one is required : it seems **pl.DataFrame** constructor handles the case where **rows** is empty for you. 
- Maybe not logging everything we currently log

In the end, and if I'm not wrong, we could have : 
```py
class MetricStructureBuilder:
    def __init__(...) -> None:
        ....

    def build(self) -> MetricStructureTable:
        rows: list[dict[str, object]] = []
        for term in self.metric.terms:
            components_ids = components_in_taxon(selmodel_library, system, term.taxonomy_category)
            for component_id in component_ids:
                component = self.system.get_component(component_id)
                breakdown_properties = _format_breakdown_properties(component.properties, self.metric.breakdown)
                rows.append(
                   {
                       "metric_id": self.metric.id,
                       "component": component_id,
                       "metric_location": metric_location,
                       "breakdown_properties": breakdown_properties,
                       "output": term.output_id,
                       "weight_output_id": 1,
                   }
                 )
        return MetricStructureTable(pl.DataFrame(rows, schema=_METRIC_STRUCTURE_SCHEMA))
```

gems_views_builder/metrics_builder.py
```diff
@@ -39,9 +40,35 @@ class MetricStructureTable:
...
+
+
+def _format_metric_location(locations: str | tuple[str, ...]) -> str:
+    locs = (locations,) if isinstance(locations, str) else locations
+    if not locs:
+        return "{}"
+    return "{" + ",".join(locs) + "}"
 
+
+class MetricStructureBuilder:
```
guilpier-code: I'm not even sure that a builder class is required here : 
- it contains only one method
- in the end, it requires 3 data members : **model_library**, **system** and **metric**. 

These 3 pieces of data could be passed as argument of a **buildMetricStructure(...)** function.

gems_views_builder/__init__.py
dusanparipovic: Exactly as you said, when time comes library could be integrated with any third party, and this is just public API. 
You could see practical usage in tests folder where we only say:
```python
from gems_views_builder import Library 
```
not 
```python
from gems_views_builder.input.library import Library 
```
Practical usage shorter import

gems_views_builder/input/system.py
dusanparipovic: Nice remark I thought this is possible and we expect to return tuple (component2,component3,component2)  in fact we need to raise error because component2 appeared 2x ?
```yml
component1 <-port1-> component2 
component1 <-port1-> component3 
component1 <-portN-> component2
``` 
Component1 is connected to component2 via multiple ports and we expect to return (component2,component3,component2) 
In fact I think this is possible to write in GEMS without exception ? Correct me if I'm wrong but we should then prevent it from side of `gemspy`.

gems_views_builder/metrics_builder.py
dusanparipovic: Simplified in #66

gems_views_builder/input/system.py
dusanparipovic: @tbittar

dusanparipovic: @aoustry @tbittar Could you answer on this before I push code ? 
One more remark from my side, this could be possible but if we define dedicated equations for that in library ? 
I think we need to have check for it inside `gemspy` if that is possible.

aoustry: @dusanparipovic As explained by another channel,

- For sure in GemsPy or Antares Simulator, a component can be connected to several other components through the same port (ex: area connected to several generators and loads through balance_port)
- However in the context of ViewsBuilder, for a given port used as location for a Metric, a component should have a unique peer through this port. Otherwise, the notion of "location" makes no sense.
- It is true that there might be several location ports : for each port, the peer component should be unique.

gems_views_builder/__init__.py
guilpier-code: ok, so it would be kind of a public interface for **gems_views_builder**.
Normally, unless the need is expected within a short time, it should be avoided to add things we don't need for sure.
It bloats the code base unnecessarily. 
We should wait for the need of a piece of code before adding it to code base.
Leave it suits you.

tests/test_input_system.py
aoustry: @dusanparipovic is this comment taken into account somewhere ?

tests/test_model_library.py
```diff
@@ -51,16 +51,28 @@ def test_model_library_models_are_typed(test_dataset_dir: Path) -> None:
...
 
 
+# Known model id -> taxonomy category for datasets that include that model.
+_KNOWN_TAXONOMY_CATEGORIES: dict[str, str] = {
+    "area": "balance",
+    "bus": "balance",
+    "generator": "production",
+    "generator_basic": "production",
+    "load": "consumption",
+    "store": "consumption",
```
aoustry:
```suggestion
    "store": "storage",
```

tests/test_input_system.py
dusanparipovic: Yes, check below in `test_input_system.py` 
`def test_locating_function_zero_peers_raises(test_dataset_dir: Path) -> None:`

dusanparipovic: I added one more additionally @aoustry

gems_views_builder/input/catalog.py
guilpier-code, line 1: **_Out of scope of this PR (but I don't know where to put this remark otherwise)_** : 

Let's look at class **ViewBuilder** : 
```python
class ViewBuilder:
    def __init__(self, input_data: InputData) -> None:
        ...
    def build(self) -> list[MetricView]:
        metric_views: list[MetricView] = []
        for catalog_id, metric_ids in self.input_data.view_config.catalog_to_metrics.items():
            catalog = self.input_data.catalogs[catalog_id]
            for metric_id in metric_ids:
                try:
                    metric = catalog.get_metric(metric_id)
                except ValueError:
                    logging.warning(f"Metric {metric_id} not found in catalog {catalog_id}")
                    continue
                ...
```
As a view config (see **view_config.yml**) contains a list of metrics definitions, we expect, at this stage, to loop over a list of metrics only, without a referencing the catalog each metrics belongs to.
So we could expect having something like : 
```python
class ViewBuilder:
    def __init__(self, input_data: InputData) -> None:
        ...
    def build(self) -> list[MetricView]:
        metric_views: list[MetricView] = []
        for metric in self.input_data.view_config.metrics.items():            
                ...
```
much more simple, and natural because it corresponds better to specifications : 
`LOOP over metrics listed in the view: for each metric listed in the view definition file....`


Except if I miss something, storing a correct list of **metrics** in the **view config** should be done before even creating a **ViewBuilder** object, but after input data is loaded. 

Still if I'm right, doing this would also simplify class **InputData**, as we no longer need it to contain a list of catalogs.

To do this, we have to load catalogs first, then load view config, then fetch the metrics it contains in the list of catalogs. So I wouldn't be surprised to see : 
```python
class ViewConfig:
    id: str
    input_data_path: Path
    calendar_id: str
    location_taxonomy_category: str | None = None
    catalog_ids: list[str] = field(default_factory=list)
    time_aggregation: TimeAggregation | None = None
    metrics: list[Metric]
   
    def fetchMetrics(catalogs: list[Catalog]):
        ....
```

gems_views_builder/metrics_structure_builder.py
```diff
@@ -51,19 +77,28 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
-                            "weight_output_id": 1,
-                        }
-                    )
+                    component = self.system.get_component(component_id)
+
+                    # # Decide does the component matches the filter, if yes they will contribute to the metric
+                    if _check_filter_matches(component, metric.filter):
+                        metric_location = _format_metric_location(
+                            self.system.get_location(component_id, term.location_ports)
+                        )
```
aoustry, line 101: Please add the following check: the code should do an assert on `metric_location` is a component whose model belong to the taxonomy-category declared as ViewConfig.location_taxonomy_category.

aoustry (changes requested):

gems_views_builder/metrics_structure_builder.py
guilpier-code, line 69: About `def build(self, metric: Metric)` : 
This method contains 3 nested loops and an **if** statement.
It makes the method long and complex, and thus hard to read.

Looking deeper to what it does, here's what I would say : 
- to avoid a loop, we could imagine a short cut, at the condition that input data are correctly formed and prepared (especially the system components). 
- this function is not the place to make checks such as : for a given component, at the other end of a port field, do we have the same taxonomy category ? To me, this should have been done earlier : before creating a **MetricStructureTableBuilder** or even a **ViewBuilder**, in a step where we prepare and connect some data we need : each components should know few things about itself at this stage, it should be able to answer questions such as : do I match a filter ?  Who am I connected to (component, port) ? Components should not discover answers to these questions  when building a **MetricStructureTable**.

If you ask me : "how would be the ideal structure for this **build()** method ?", I would answer (I skip **logging** for readability) : 
```python
def build(self, metric: Metric) -> MetricStructureTable:
    rows: list[dict[str, object]] = []
    for term in metric.terms:
        for c in self.components if c.match(term.taxonomy_category) and c.match(metric.filter) :
            rows.append(
                {
                    "metric_id": metric.id,
                    "component": c.id(),
                    "metric_location": c.formatted_locations(term.taxonomy_category),
                    "breakdown_properties": c.matching_breakdown(metric.breakdown),
                    "output": term.output_id,
                    "weight_output_id": 1,
                }
           )
    return MetricStructureTable(rows, metric)
```
Note that : 
- code is simpler, shorter and natural (in particular, a **for** loop disappeared). It sticks to the specifications (**2.1.1**).
- It assumes that a **MetricStructureTableBuilder** object is built from (and contains) the list of system components
- It assumes that components know things about themselves (which is understandable, is it not ?) 
- Checks were removed, because building a **MetricStructureTable** is not the place for this.
- metric locations (retrieved from components) should be a tuple (general case) and not a choice of types, for simplicity

So, in the end, we could have this plan : 
- read / load input data
- connect data : we should complete components with pieces of information on themselves from elsewhere : do I match a filter ? A taxonomy category (= **taxon** [yes it does exist]) ?  Breakdown properties ? What are my locations in relation with a taxonomy category ?
- build a **ViewBuilder**
- print the final view

gems_views_builder/metrics_structure_builder.py
```diff
@@ -51,19 +77,28 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
+                                "weight_output_id": 1,
+                            }
+                        )
+                    else:
+                        logging.debug(
+                            f"[{metric.id}] Component {component_id!r} did not match metric filter and was skipped"
+                        )
 
         rows = pl.DataFrame(rows_data, schema=METRIC_STRUCTURE_TABLE_SCHEMA)
         return MetricStructureTable(rows, metric.id)
```
guilpier-code, line 118: _**Making code a bit simpler**_
```python
def build(self, metric: Metric) -> MetricStructureTable:
    rows_data: list[dict[str, object]] = []
    ...
    rows = pl.DataFrame(rows_data, schema=METRIC_STRUCTURE_TABLE_SCHEMA)
    return MetricStructureTable(rows, metric.id)
```
At this level : 
- we don't want to know that a **MetricStructureTable** is actually a data frame under the hood. It's an implementation detail. We should make it a data frame inside the **MetricStructureTable** constructor.
- we could pass a metric (not a metric id)

```python
def build(self, metric: Metric) -> MetricStructureTable:
    rows: list[dict[str, object]] = []
    ...
    return MetricStructureTable(rows, metric)
```
Makes the code a bit simpler and natural.

gems_views_builder/metrics_structure_builder.py
guilpier-code, line 76:
```python
logging.debug(
    f"[{metric.id}] Processing term for taxonomy category {term.taxonomy_category!r} "
    f"and output {term.output_id!r}"
)
```
It could be shorter. 
Not sure we need all these words and details. For example, do we need the term's **taxonomy category** ?
```python
logging.debug(f"   Processing term : {term.output_id!r}")
```

gems_views_builder/metrics_structure_builder.py
```diff
@@ -51,19 +77,28 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
-                            "component": component_id,
-                            "metric_location": loc_str,
-                            "breakdown_properties": "",
-                            "output": term.output_id,
-                            "weight_output_id": 1,
-                        }
-                    )
+                    component = self.system.get_component(component_id)
+
+                    # # Decide does the component matches the filter, if yes they will contribute to the metric
```
guilpier-code:
```python
# # Decide does the component matches the filter, if yes they will contribute to the metric
```
Useless comment : most of the time, we don't need comments.
Expressive code makes comments useless.
There are cases where comments can can be useful, but not here.

gems_views_builder/input/catalog.py
dusanparipovic: Right make sense, I thought that is more natural to process catalog by catalog but in the end we're fine.

gems_views_builder/input/system.py
```diff
@@ -31,104 +37,146 @@ class System:
...
+            raise ValueError(
+                f"Expected exactly one peer component for component {component_0_id!r} "
+                f"on port {location_port!r}, but found {len(peers)}: {peers!r}"
+            )
+        peer = peers[0]
+        logging.debug(f"Resolved location for component {component_0_id!r} via port {location_port!r} to {peer!r}")
+        return peer
 
 
 def load_system(input_data_path: Path) -> System:
```
guilpier-code: **_Unfinished comment, need to have more info on GemsPy to finish_**

Here we make references to **C++** software **Antare_Simulator**, which also loads libraries and systems. 
We supply links to its git repo when useful, especially in this [source folder](https://github.com/AntaresSimulatorTeam/Antares_Simulator/tree/develop/src/modeler/loadFiles).

`def load_system(input_data_path: Path) -> System:` : 
Here loading system implies loading libraries as well.
This is misleading.
We should separate things (see [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/develop/src/modeler/loadFiles/loadAll.cpp)) : 
- loading librairies : returns a (list of ?) **Library** (each Library contains a list of **Model**)
- loading system : takes a **Library** as argument and returns a list of **Component** (each **Component** references the **Model** it's an instance of)
- About connections between 2 components through a port field : every component should contain the list of components it's connected to through a port field (via a dictionary : local port -> list of remote (Component, Port)).

In relation with [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/51#discussion_r3511504671), main program could contain : 
```python
...
input_data = Loader(input_dir).load()
input_data.system.updatewith(input_data.library)
...
```

gems_views_builder/input/catalog.py
guilpier-code, line 86: `location_ports: str | tuple[str, ...] | None`
Why does **location_ports** has a so complex type ?
If we make it a **tuple**, then either : 
- it is empty --> replaces **None**
- it contains a unique element --> replaces --> **str**
- it contains several elements

gems_views_builder/metrics_structure_builder.py
dusanparipovic: We don't need Taxonomy category for now but in close future it will be needed. 
If you closely look at the diagram workflow Taxonomy is also included process of building MetricStructureTable.
Once we include [this](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/54), it will be good to have this log for my opinion. Log will be changed but point is same. 
It will be something like:  Processing metric id term for taxonomy category which belongs to/which is root category.

gems_views_builder/view/views_builder.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, input_data: InputData) -> None:
         self.input_data = input_data
         # # Builder which is reusable over metrics
         self.metric_structure_table_builder = MetricStructureTableBuilder(
-            self.input_data.system, self.input_data.library
+            self.input_data.system, self.input_data.library, self.input_data.view_config.location_taxonomy_category
```
guilpier-code, line 28:
```python
self.metric_structure_table_builder = MetricStructureTableBuilder(
    self.input_data.system, self.input_data.library, self.input_data.view_config.location_taxonomy_category
)
```
In relation with [this main comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/51#discussion_r3511504671), we could only need `self.input_data.system.components`
to build a **MetricStructureTableBuilder** object.
```python
self.metric_structure_table_builder = MetricStructureTableBuilder(self.input_data.system.components)
```
We could even have **MetricStructureTableBuilder** created outside **ViewBuilder** and passed to **ViewBuilder**'s constructor (your choice).

gems_views_builder/input/system.py
```diff
@@ -31,104 +37,146 @@ class System:
 
     def __init__(self, system: GemsSystem) -> None:
         self._system = system
```
guilpier-code, line 37: ` self._system = system` : _comment to be finished_

gems_views_builder/input/catalog.py
dusanparipovic: None - practically means we're not required to write it 
str - one location port 
tuple - multiple ports, I would say this is prevention from changing over program flow 
How we can simplify this:
`location_ports: list[str] | None` 
@aoustry  ?

gems_views_builder/input/system.py
```diff
@@ -31,104 +37,146 @@ class System:
...
-        self._components_by_model: dict[str, list[str]] = self.models_to_components()
+        logging.info(
+            f"Initializing input system with {len(self.components)} component(s) "
+            f"and {len(self.connections)} connection(s)"
+        )
+        self._components_by_model: dict[str, list[str]] = self._models_to_components()
         self._component_port_connections: dict[tuple[str, str], set[str]] = self.build_component_port_connections()
```
guilpier-code, line 43: According to [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/51#discussion_r3511504671), we may not need these 2 dictionaries : each component is supposed to know (and be supplied with) things about itself. 
So these pieces of information are contained in each component.

gems_views_builder/input/system.py
```diff
@@ -31,104 +37,146 @@ class System:
...
+        )
 
     @property
-    def components(self) -> list[GemsComponent]:
-        return cast(list[GemsComponent], self._system.components)
+    def components(self) -> list[Component]:
+        return list(self._system.components)
 
     @property
     def connections(self) -> list[Any]:
```
guilpier-code, line 54: For same reason, we should not need this (connections inside components)

gems_views_builder/input/system.py
```diff
@@ -17,11 +17,17 @@
 from pathlib import Path
 from typing import Any, cast
 
-from gems.study.parsing import (  # type: ignore
-    ComponentSchema as GemsComponent,
+from gems.model.resolve_library import resolve_library  # type: ignore
+from gems.study import Component  # type: ignore
```
guilpier-code, line 21: I just saw this importation from **GemsPy**.
There's a discussion to have on this about class **Component** (for instance).
In this dependency, should **Component** hold a list of connections to other component ?
This question is related to [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/51#discussion_r3511504671)

gems_views_builder/metric_structure_table.py
```diff
@@ -5,24 +5,30 @@
...
     # # One edge case here is that directories never be cleaned up, only will be cleaned while calling destructor
-    def __init__(self, rows: pl.DataFrame, metric_id: str) -> None:
+    def __init__(self, rows: list[dict[str, object]], metric_id: str) -> None:
         self._tmp_root = Path(tempfile.mkdtemp())
-        if rows.is_empty():
+        if len(rows) == 0:
             logging.info(f"[{metric_id}] No matching components found — metric structure table is empty")
         else:
-            logging.info(f"[{metric_id}] Metric structure table built with {len(rows)} row(s)")
+            logging.info(f"[{metric_id}] Metric structure table built with {len(rows)} rows")
```
guilpier-code: Or simply : 
```python 
logging.info(f"[{metric_id}] Metric structure table has {len(rows)} rows")
```
whatever the number of lines : can be zero.

tests/test_business_view_config.py
```diff
@@ -49,9 +49,14 @@ def test_view_config_known_values(test_dataset_dir: Path) -> None:
     assert config.id == "view_area"
```
guilpier-code, line 51: All test start with **test_view_config_**.
It's not mandatory, as we are in file whose name clearly tells its subject.

tests/test_view_builder.py
guilpier-code:
```python
accumulate_on_disk(metric_views, results_dir)
```
We should have the choice on the format we want to print the view into (**parquet** ? **csv** ?)
So, at some point, we should have another print function into csv.

gems_views_builder/input/catalog.py
dusanparipovic: ### Important Remark

The view configuration must be loaded **before** loading any catalogs, since it defines which catalogs are required.

The workflow should therefore be:

1. Load the view configuration.
2. Extract the referenced catalog IDs.
3. Load only the required catalogs.
4. Use those catalogs to resolve and populate the referenced metrics in the view configuration.

This design will remain valid even after we remove the current folder-based catalog structure. Regardless of how catalogs are stored in the future, the view configuration will always be the source of truth for determining which catalogs need to be loaded.

Another important consideration is that, in the near future, we will support **multiple view configurations** running on the same input data. This makes loading catalogs based on the view configuration even more important.

Although a catalog may define many metrics, a particular view configuration typically references only a subset of them. Therefore, the pipeline should resolve only the metrics explicitly requested by the active view configuration(s).

That said, I agree with your idea. During the build process, we should iterate over the resolved metrics rather than the catalogs themselves. Once the catalogs have been loaded and the metric references resolved, the rest of the pipeline should operate purely on metrics without needing to be aware of the underlying catalog structure.

Example:

```yaml
catalog:
  # References the business metric definition files.
  - id: catalog

metrics:
  # References metrics defined in the selected catalog.
  - id: catalog.PROD
  - id: catalog.LOAD
  - id: catalog.BALANCE
```

gems_views_builder/input/system.py
dusanparipovic: I agree with all comments but then we need to perform update inside `gemspy` if we want to achieve that. 
As you can see this is way how we handle until we create GEMS Craft library. 
I don't know do we see interest to integrate this now inside gemspy? @aoustry @tbittar ? 
Maybe I could write issue ?

gems_views_builder/metrics_structure_builder.py
dusanparipovic: Very good approach, I agree but now questions are:

1. Do we want to extend component inside gemspy or to override component inside gvb
2. Do we want to apply this changes inside this PR

gems_views_builder/metrics_structure_builder.py
dusanparipovic: Done!

tests/test_business_view_config.py
dusanparipovic: Right but we need to keep test_ because that's how `pytest` library internally collect all tests.

tests/test_view_builder.py
dusanparipovic: Done [here](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/77/changes).

dusanparipovic: Also as we talked I added support that simulation table input could be csv.

gems_views_builder/input/view_config.py
guilpier-code: `parsed = load_view_file(config_file_path)` : 

- we could rename **parsed** into **raw_view_config** (or **raw_vw_cfg** maybe), more clear.
- In the same idea, **MetricRef** could be renamed into **MetricId** (same for **catalogs**).
- **metric_references** ==> **metric_ids**
- **ViewData** ==> **RawViewConfig**
- **load_view_file(...)** ==> **load_raw_view_config(...)**

gems_views_builder/input/view_config.py
```diff
@@ -88,18 +95,19 @@ def load_view_config(config_file_path: Path) -> ViewConfig:
...
     view_config = ViewConfig(
         id=parsed.id,
         input_data_path=input_data_path,
         calendar_id=calendar_id,
         location_taxonomy_category=location_taxonomy_category,
-        catalog_ids=[c.id for c in parsed.catalog],
+        catalog_ids=catalog_ids,
         time_aggregation=parsed.aggregation[0].time if parsed.aggregation else None,
-        catalog_to_metrics=group_metrics_by_catalog(parsed.id, parsed.metrics),
+        metric_ids_by_catalog=group_metrics_by_catalog(catalog_ids, parsed.metrics),
```
guilpier-code:
```python
metric_ids_by_catalog=group_metrics_by_catalog(catalog_ids, parsed.metrics),
```
**metric_ids_by_catalog** is not necessary (and not natural) as a data member of **ViewConfig**.
Actually it's useful only when we call method **ViewConfig.fetch_metrics(...)**.
Instead, **metric_ids** and **metrics** (as a **ViewConfig**) are expected;
So first, I would do : 
```python
view_config = ViewConfig(
        id=raw_view_config.id,
        input_data_path=input_data_path,
        calendar_id=calendar_id,
        location_taxonomy_category=location_taxonomy_category,
        catalog_ids=catalog_ids,
        time_aggregation=raw_view_config.aggregation[0].time if raw_view_config.aggregation else None,
        metric_ids = [id for id in raw_view_config.metrics] # or something like that
        metrics = [] # or whatever list
```
Then I would make **group_metrics_by_catalog** as private method of **ViewConfig** and we could have : 
```python
class ViewConfig:
    ...

    def fetch_metrics(self, catalogs: dict[str, Catalog]) -> None:
        metric_ids_by_catalog = _group_metrics_by_catalog()
        for catalog_id in metric_ids_by_catalog.keys():
            for metric_id in metric_ids_by_catalog[catalog_id]:
                self.metrics.append(catalogs[catalog_id].get_metric(metric_id))

    # I am private, no arg needed
    def _group_metrics_by_catalog() -> dict[str, set[str]]:
        metric_ids_by_catalog: dict[str, set[str]] = defaultdict(set)
        # Having access to self.catalog_ids and to self.metric_ids
        return metric_ids_by_catalog
```

Note that **metric_ids_by_catalog** is currently used in **test_business_view_config.py**.
As it would disappear, 1 or 2 tests would have to be slightly changed.

gems_views_builder/input/view_config.py
```diff
@@ -63,67 +64,75 @@ class ViewConfig:
...
         calendar_id=calendar_id,
         location_taxonomy_category=location_taxonomy_category,
-        catalog_ids=[c.id for c in parsed.catalog],
-        time_aggregation=parsed.aggregation[0].time if parsed.aggregation else None,
-        catalog_to_metrics=group_metrics_by_catalog(parsed.id, parsed.metrics),
+        catalog_ids={c.id for c in raw_view_config.catalog},
+        time_aggregation=raw_view_config.aggregation[0].time if raw_view_config.aggregation else None,
+        metric_ids=[metric.id for metric in raw_view_config.metrics],
     )
+    view_config._group_metrics_by_catalog()
```
guilpier-code: Why do we call `view_config._group_metrics_by_catalog()` here if we call it inside **fetch_metrics** already ?

gems_views_builder/input/system.py
aoustry: @dusanparipovic for now I think that this is easier that you extend **_in the GVB code_** the GemsPy classes you're using, buy adding the methods you need. In the future, we may push this code in GemsPy/GemsCraft.

aoustry (dismissed):

gems_views_builder/input/catalog.py
dusanparipovic: FYI we agreed that we will keep `location_ports: str | tuple[str, ...] | None` maybe it looks tiny complicated but it won' force user to write list with one element.

PR #52, opened 2026-06-04
feat(spatial-aggregation): aggregate metrics by location-component property

### LINKED ISSUE #16: feat: spatial aggregation
Implementing a concept equivalent to the "district" concept in AntaresLegacy.

The approach would consist in relying on the "properties" of the Components belonging to the Taxonomy Category used as location.

tbittar:
09.06.26: PLEASE IGNORE THIS PR WHILE THE PRs #49, #50, #51 are not merged

## Summary

- `metrics.py`: add `LocationAggregation(key: str, on_missing: 'keep'|'drop' = 'keep')` Pydantic model; add `location` field to `Aggregation`; `ViewConfig` reads and exposes `location_aggregation`
- `metrics_builder.py`: add `_resolve_location_aggregation()` — relabels each raw location component ID to the value of the configured property key; `on_missing='keep'` emits `<unknown>`, `on_missing='drop'` excludes the row entirely; `MetricStructureBuilder` accepts an optional `location_aggregation` parameter
- `views.py`: passes `view_config.location_aggregation` through to `MetricStructureBuilder`

## Test plan

- [ ] `test_view_builder.py` — end-to-end test with `test_location_aggregation` dataset passes
- [ ] `test_location_aggregation/view_config.yml` (keep) and `view_config_drop.yml` (drop) both produce correct aggregated outputs
- [ ] `test_location_aggregation/view_config_no_location.yml` verifies no-aggregation path is unchanged


resources/tests_inputs/test_3/catalogs/catalog.yml
```diff
@@ -0,0 +1,67 @@
...
+  metrics-definition:
+
+  - id: PROD
+    terms:
+      - taxonomy-category: production
+        output-id: p
+        location-ports: p_balance_port
+    terms-operator: sum
+    time-operator: sum
+    breakdown-property:
```
tbittar: **Stale field name**: The catalog still uses `breakdown-property:` (old singular key) while the new `MetricData` model has `breakdown:` (a list). Because `ViewBuilderBasedModel` uses Pydantic's default `extra='ignore'`, this key is silently dropped and breakdown is `None`. Tests pass only because test_3 never asserts on breakdown output. Update to `breakdown: [{key: business_view}]` or remove it entirely if breakdown is not needed for this test fixture.

tbittar (commented):

dusanparipovic:
I think everything is resolved but I need to do one more check after holidays. In fact Claude didn't performed very well in PR refactoring.

gems_views_builder/metrics_structure_builder.py
```diff
@@ -81,8 +105,19 @@ def build(self, metric: Metric) -> MetricStructureTable:
 
                     # # Decide does the component matches the filter, if yes they will contribute to the metric
                     if _check_filter_matches(component, metric.filter):
-                        metric_location = _format_metric_location(
-                            self.system.get_location(component_id, term.location_ports)
+                        raw_location = self.system.get_location(component_id, term.location_ports)
+                        raw_locations = [raw_location] if isinstance(raw_location, str) else list(raw_location)
+                        resolved_locations = self._resolve_location_aggregation(raw_locations)
+                        if not resolved_locations:
```
aoustry: you mean if len(resolved_locations) == 0 ?

gems_views_builder/metrics_structure_builder.py
```diff
@@ -54,9 +55,32 @@ def __init__(
...
+        if cfg is None:
+            return locations
+        result: list[str] = []
+        for loc in locations:
+            val = self.system.get_component(loc).properties.get(cfg.key)
+            if val is not None:
+                result.append(val)
+            elif cfg.on_missing == "keep":
+                result.append("<unknown>")
+            # else on_missing == "drop": skip this location
```
aoustry: FIX NEEDED, this is not the expected behavior for the "drop" mode in the case 

in the drop mode, if one of the loc in locations has a None value for `val`, then there should NOT be an entry in the MetricStructureTable.

That is why I would suggest to add 
```
elif cfg.on_missing == "drop"
       return []
else: #raise an Error 
```

aoustry: @dusanparipovic

gems_views_builder/metrics_structure_builder.py
aoustry: @dusanparipovic

gems_views_builder/metrics_structure_builder.py
dusanparipovic: Done. 
One remark there's no need for raising error. 
Why because `pydantic` will automatically raise an error if `location_aggregation` is not in {drop,keep}.

gems_views_builder/metrics_structure_builder.py
dusanparipovic: Yes that's basically same if statement because, `resolve_location_aggregation` will return list anyways, so if we have empty list as return result it will be treated as false. This is very `pythonic` syntax.

aoustry:
Since we reshaped the design of this feature, I would suggest a new start, from scratch

Issue #62, opened 2026-06-15
fix(view config): file references


dusanparipovic:
# Current state
```yml
view:

  id: view_area

  scope:
    - location:
      taxonomy-category: balance

    - calendar: calendar_file

  aggregation:
    - time: hour
    - location:
        key: country
        on_missing: keep

  catalog:
    - id: catalog

  metrics:
    - id: catalog.LOAD
    - id: catalog.PRODUCTION
    - id: catalog.BALANCE
```
# Proposal
```yml
view:

  id: view_area

  scope:
    - location:
      taxonomy-category: balance

    - calendar: path_to_calendar_file
      alias: calendar_alias  # I'm not sure that is needed to we want to be able in future to use multiple calendars inside workflow
    - simulation_table: path_to_simulation_table 
      alias: sim_table_alias # same question 
    - library: path_to_library 
    - taxonomy: path_to_taxonomy 
    - system: path_to_system

  aggregation:
    - time: hour
    - location:
        key: country
        on_missing: keep

  catalog:
    - id: path_to_catalog_file_one
      alias:  catalog
    - id: path_to_catalog_file_two
      alias: alias_for_catalog_two

  metrics:
    - id: catalog.LOAD
    - id: catalog.PRODUCTION
    - id: catalog.BALANCE
    - id: alias_for_catalog_two.SOMETHING
```

dusanparipovic:
With this approach we will have single entry point to our library.
User will just pass view config file which will reference all other files needed for workflow. 
With this we don't need to have specific folder structure as we have required folder for catalogs and all other files placed inside one folder.

tbittar:
The view config file has the responsability of defining the metrics that are built within a view. With a given simulation output we can have have several view config file. The configuration of the folder structure and the view config are not at the same level, they should be configured in separate file (1 folder structure config and 1..N view config files)

PR #66, opened 2026-06-17
refactor(chore): general improvement of code quality based on comments from multiple PRs 

### LINKED ISSUE #61: [GEMSPY UPDATE] New release: v0.1.2
> This issue was automatically created by the `notify-gemspy-release` workflow in GemsPy.

## New Version Detected

| | |
|---|---|
| **New version** | `0.1.2` |
| **PyPI page** | https://pypi.org/project/gemspy/0.1.2/ |
| **Release notes** | https://github.com/AntaresSimulatorTeam/GemsPy/releases/tag/v0.1.2 |

## Triage

- [ ] GemsPy changelog reviewed
- [ ] `gemspy` dependency version updated in `pyproject.toml` (if new features or fixes are required)
- [ ] No action needed (if release has no impact on GEMS-ViewsBuilder)

### LINKED ISSUE #65: refactor(core): refactor project structure and isolate class responsibility
- [x] Make new directory `input` which will be place for core classes e.g. input files(taxonomy,system,simulation table...) 
- [x] Isolate class responsibility 
- [x] Currently each class has own load inside class that needs to be moved out for easier testing 
- [x] ViewBuilder is responsible for things like merging results,cleaning intermediate files,etc... 
 Idea is to move that out of ViewBuilder 
- [x] This logic needs to be applied over repository and to be aligned based on Guillaume comments #49  (single responsibility, better for testing)
- [x] Add missing license header
- [x] Restructuring of ViewBuilder - clean logic, nothing is loaded inside ViewBuilder but we will pass new class InputData
- [x] Restructuring of Metric Structure Table - same as ViewBuilder
- [x] Logs should be placed inside classes not inside ViewBuilder, this class needs to be as much clean

### LINKED ISSUE #67: feat(CLI): introduce command line interface
Introduce command line interface as current main program

dusanparipovic:


gems_views_builder/views_builder.py
```diff
@@ -0,0 +1,46 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+"""ViewBuilder."""
+
+from gems_views_builder.aggregator import Aggregator
```
guilpier-code: **ViewBuider** is now much cleaner.
That said, it could be even cleaner : this algorithm should not see / know about : 
- input_data_path
- writer

Why ? Because this algo is only about building a view, nothing else. **input_data_path** or **writer** imply we mix this algo with something external (loading and writing) or that we don't hide implementation details.

So my step-by-step recommendations would be : 
- Get rid of the **input path** in **aggregator** : we should build the aggregator outside this algo and before creating it : 
```python
class ViewBuilder:
    def __init__(self, input_data: InputData, aggregator : Aggregator) -> None:
          self.input_data = input_data
          self.aggregator  = aggregator
    
    def build(self) -> MergedView:
        ...
        return merged

# Now building the ViewBuilder : 
input_data = ... # building the input
aggregator = Aggregator(input_data_path)
viewBuilder = ViewBuilder(input_data, aggregator)
mergedView = viewBuilder.build()
```
   **Caution** : regarding **aggregator**, there is an update in a [later remark](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/66#discussion_r3442826567)
- Get rid of the **writer** in **ViewBuilder** (in gereral): writer currently exists in class **ViewBuilder** for persistence reason. 
  But we don't want to know about persistence at this level of code because it's a technical consideration.
  We chose to persist objects of the algorithm on disk for RAM reason, but we could have chosen another solution.
  The algorithm of building the a view doesn't have to know these technical details.
  The expected benefits are :
   - Simplification of class **ViewBuilder**
   - If we change persistence policy or other details, class **ViewBuilder** remains unchanged.
  
  To remove this writer, I suggest that each object that has to be saved on disk handles its persistence by itself :
  It should create the file on disk at construction and remove that file at destruction (`def __del__`).
  So the idea is that an object encapsulates the persistence and minds its own business on that matter.
  No one else than itself should know that this object needs to persist on disk.
  
- Get rid of the **writer** in **build_metric_structure** : looking at this function, I see strange things.
   - class **MetricStructureBuilder** is supposed to build a **MetricStructure**, not a **MetricStructureTable** (which is almost empty and does nothing by the way). Let's not do that, let's make **MetricStructureBuilder** build a **MetricStructure** that encapsulates its own data frame and its own path for persistence (please see below)
   - **build_metric_structure** creates a **MetricStructureBuilder** and then calls its **build()** method. If we do as suggested above, **build_metric_structure** becomes useless.
   - Instead of **build_metric_structure**, create a **MetricStructureBuilder** outside of the view builder, and pass it to the **ViewBuilder** constructor. At **MetricStructureBuilder** creation, pass directly the **input_data_path** and also **system** and **library** (none of them change during **ViewBuilder** process).
   - The code should now look like :
```python
class ViewBuilder:
    def __init__(self, input_data: InputData, 
                 aggregator : Aggregator, 
                 metricStructureBuilder : MetricStructureBuilder) -> None:
          self.input_data = input_data
          self.aggregator  = aggregator
          self.metricStructureBuilder = metricStructureBuilder
    
    def build(self) -> MergedView:
        ...
        for catalog_id, metric_ids in ...
             ...
             for metric_id in metric_ids:
                 ...
                 metricStructure = metricStructureBuilder.build(metric)
                 metric_views.append(..., metricStructure, metric)
        ...
        return merged

# Now building the ViewBuilder : 
input_data = ... # building the input
aggregator = Aggregator(input_data_path)
metricStructureBuilder = MetricStructureBuilder(system, library, input_path)
viewBuilder = ViewBuilder(input_data, aggregator, metricStructureBuilder)
mergedView = viewBuilder.build()
```
- Get rid of the **writer** when merging views : looking at `MergedView.merge_views`, I see that `writer.merge_results` is called. So the writer contributes to merging object in a way.
     I'm not confortable with this idea (a writer should write only)
     Code exerpt in this remark wasn't that bad actually.
```python
MergedView view()
for catalog_id, metrics in ...
    ...
    for metric_id in metrics:
        ...
        view.add(metric_view)
     view.merge()
     return view
```
   Here **MergedView** encapsulates merging, which involves persistence. The result of creating **MergedView** seems to be final view. So this class does not have to have a destructor.

gems_views_builder/views_builder.py
```diff
@@ -0,0 +1,46 @@
...
+                    continue
+                structure = build_metric_structure(
+                    self.input_data.system,
+                    metric,
+                    self.input_data.library,
+                    self.writer,
+                )
+                metric_views.append(self.aggregator.aggregate(self.input_data.filtered_st, structure, metric))
+        merged = MergedView.merge_views(metric_views, self.writer)
+        self.input_data.filtered_st.cleanup()
```
guilpier-code:
```python
self.input_data.filtered_st.cleanup()
```
This shouldn't be done here, it's a technical detail that has nothing to see with building views.
This could be done in class **FilteredSimulationTable**'s destructor.

Indeed, we could feel like destroying this object earlier than it would normally be destroyed (end of the scope it was created in), when we're sure that it will no longer be used. But in that case, the **cleanup()** method is called at the end of the view building algorithm, when filtered ST is no longer useful.

gems_views_builder/writer.py
```diff
@@ -1,16 +1,32 @@
...
+# See AUTHORS.txt
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
```
guilpier-code: As a consequence of the **gems_views_builder/views_builder.py** review, the class writer disappears and is kind of spread into classes **MetricStructure** and **MergedView** (for the persistence part).
This is not a bad thing because the writer currently writes 2 heterogeneous things, and thus is not consistent.
That's why I recommend spreading its code to destination classes were there is a real consistency.

This source file (**writer.py**) should then be renamed into **MergedView.py** if the recommendations to remove the **writer** are followed.

gems_views_builder/writer.py
```diff
@@ -1,16 +1,32 @@
...
+import logging
```
guilpier-code: As a consequence of the **gems_views_builder/views_builder.py** review, class **MergedView** gets 2 methods **add(...)** and **merge()**.
The current method **merge_views** disappears or is replaced with **merge()**

gems_views_builder/metric_view.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# SPDX-License-Identifier: MPL-2.0
+#
+# This file is part of the Antares project.
+
+from dataclasses import dataclass
```
guilpier-code, line 14: As a consequence of **gems_views_builder/views_builder.py** review, class **MetricView** gets a constructor that takes an output path as argument, for persistence. As creating a **MetricView** object consists in (or is not separable from) creating it on disk, the constructor should do it.
A possible destructor should delete it from disk.

gems_views_builder/metric_view.py
```diff
@@ -0,0 +1,24 @@
...
+from pathlib import Path
+
+
+@dataclass
+class MetricView:
+    """Temporal aggregation result for a single metric, stored as a parquet file."""
+
+    file: Path
+
+    def cleanup(self) -> None:
```
guilpier-code: What **cleanup** does seems actually the mission of a **destructor**.
As you know, the advantage of a class **destructor** is that it's called automatically, no need to call it explicitly.

gems_views_builder/metrics_builder.py
```diff
@@ -12,13 +12,14 @@
 
 import logging
```
guilpier-code: According to **gems_views_builder/views_builder.py** review : 
- class **MetricStructureTable** should be removed because is contains so little (and seems useless). Actually, we have currently 2 classes **MetricStructureTable** and **MetricStructure**, but only one should remain. Whatever the name, the content and behavior is the same (**specifications** talk about **MetricStructureTable**, so maybe this name should be chosen).
- Let's assume that **MetricStructure** is chosen
- class **MetricStructure** gets everything to possibly store the content of a metric structure (dataframe), but more likely to persist on disk (output path for ex). Its constructor prints on disk for persistence. From what I can see from **MetricStructureBuilder**, a **MetricStructure** should be created from **rows** and output path (as arguments).
- class **MetricStructureBuilder** builds **MetricStructure**, and **build** method creates and returns a **MetricStructure**. By the way, build does not need to worry about the content of **rows** (empty or not), it's handled by class **MetricStructure** itself (makes **build()** a bit shorter and cleaner).
- function build_metric_structure disappears : without class **MetricStructureTable**, without the **writer**, nothing remains. Bye.

gems_views_builder/metrics_builder.py
```diff
@@ -12,13 +12,14 @@
...
 from dataclasses import dataclass
+from pathlib import Path
```
guilpier-code: We could split this source file into 2 source files **metric_structure.py** and **metric_structure_builder.py**, one for each class

gems_views_builder/aggregator.py
```diff
@@ -24,7 +42,20 @@ def _create_metric_view_directory(self) -> Path:
...
         return metric_view_dir
 
-    def aggregate_metric_terms(
+    def aggregate(
+        self,
+        filtered_st: "FilteredSimulationTable",
+        structure: "MetricStructure",
+        metric: Metric,
+    ) -> MetricView:
+        joined = filtered_st.dataframe.join(structure.dataframe, on=["component", "output"], how="right")
```
guilpier-code:
```python
joined = filtered_st.dataframe.join(structure.dataframe, on=["component", "output"], how="right")
```
When I look at specifications, I see that step **2.b** is highlighted as an important step of the view building construction. But in **ViewBuiber's build()** method, there's no line representing this step.
It would make sense to repair this, as **build()** is supposed to be an overview of the algorithm.

How to do this ?
There are actually 2 aggregations in specifications : 
- **2.b** : terms aggregation (involves a join and group by operation)
- **2.c** : time aggregation

These are very different aggregations. I'm not sure it makes sense to aggregate everything in one unique aggregator. 
So there should be 2 aggregators actually.
As current aggregator, both new aggregators can be passed to **ViewBuilder**'s **build()** method.
But they cannot return a path (as currently do methods **_aggregate_metric_terms** or **_aggregate_metric_temporally**), because these aggregators are used in class **ViewBuilder** (no path in this class in this class should be allowed). 
They both should return a **MetricView** instead.

In the end, we may have : 
```python
class ViewBuilder:
    def __init__(self, input_data: InputData, 
                 terms_aggregator : TermsAggregator,
                 time_aggregator : TimeAggregator,
                 metricStructureBuilder : MetricStructureBuilder) -> None:
          self.input_data = input_data
          self.terms_aggregator  = terms_aggregator
          self.time_aggregator = time_aggregator
          self.metricStructureBuilder = metricStructureBuilder
    
    def build(self) -> MergedView:
        ...
        MergedView view()
        for catalog_id, metric_ids in ...
             ...
             for metric_id in metric_ids:
                 ...
                 metricStructure = metricStructureBuilder.build(metric)
                 metric_view = self.terms_aggregator.run(metricStructure, metric)
                 del metricStructure
                 metric_view = self.time_aggregator.run(metric_view, metric)
                 view.add(metric_view)
        view.merge()
        return views

# Now building the ViewBuilder : 
input_data = ... # building the input
terms_aggregator = TermsAggregator(input_data_path, input_data .filtered_st)
time_aggregator = TimeAggregator(input_data_path)
metricStructureBuilder = MetricStructureBuilder(system, library, input_path)
viewBuilder = ViewBuilder(input_data, terms_aggregator, time_aggregator, metricStructureBuilder)
mergedView = viewBuilder.build()
```
Note that : 
- When creating **TermsAggregator**, we create it with the help of **filtered ST** (and also output path, like before)
- In class **ViewBuilder**, no path is created or passed from an object to the other.
- We can delete metric structure by calling `del`, cleaner than explicit **cleanup**
- Inside **TimeAggregator's run()** method, we call **del** on passed **metric_view** and recreate a fresh one to be returned
- There are now 2 smaller aggregators, making different aggregations, and they can be called / used separately. We can change one of them without changing the other.

gems_views_builder/loader.py
```diff
@@ -1,41 +1,61 @@
...
+            input_data_path=self.input_data_path,
+            taxonomy=load_taxonomy(self.input_data_path / "taxonomy.yml"),
+            view_config=view_config,
+            catalogs=load_catalogs(self.input_data_path, view_config.catalog_ids),
+            library=load_library(self.input_data_path / "library.yml"),
+            system=load_system(self.input_data_path),
+            filtered_st=filtered_st,
+        )
+
+        validate_catalogs_against_taxonomy(input_data.catalogs, input_data.taxonomy)
```
guilpier-code: Normally, loading and validation are different task, so validating data should not be done by a loader, but after load in complete, for example by the main program.

gems_views_builder/input/simulation_table.py
```diff
@@ -0,0 +1,129 @@
...
+class FilteredSimulationTable:
+    """Lazy representation of a filtered SIMULATION_TABLE parquet file.
+
+    Has all SIMULATION_TABLE columns plus granular_date from the calendar join.
+    """
+
+    file_path: Path
+    dataframe: pl.LazyFrame
+
+    def cleanup(self) -> None:
```
guilpier-code: Having a **destructor** would be better than a **cleanup** method.
With a destructor : 
- To destroy an object intentionally, `del obj` is more natural and expected than `obj.destroy_yourself()`
- If not called explicitly, it will be called automatically

tests/test_filter_simulation_table.py
```diff
@@ -15,22 +15,21 @@
...
 import pytest
 
-from gems_views_builder import FilteredSimulationTable, SimulationTable, load_calendar
+from gems_views_builder import FilteredSimulationTable, load_calendar
+from gems_views_builder.input.simulation_table import filter_simulation_table, load_simulation_table
 
 # ---- Parametrized integration test: logical assertions (no golden overwrite) ----
 
 
 def test_filter_simulation_table_logical(tmp_path: Path, test_dataset_dir: Path) -> None:
```
guilpier-code, line 24: Here we depend on a study stored on disk at **test_dataset_dir**.
But we could make up some **calendar** and a **simulation table** in memory, using these classes constructor (class **SimulationTable** should have one by the way).
Doing this, we no longer depend on a study, and we can make light tests, as varied as possible.

gems_views_builder/input/simulation_table.py
dusanparipovic: Nice catch,thank you. 
Applied same approach for `MetricView`, rather that going over path chunks each object destructor will perform file deletion.

gems_views_builder/loader.py
dusanparipovic: I catch your point. I'll have that on mind while writing `CLI`. 
Currently this is removed from codebase.

gems_views_builder/input/catalog.py
```diff
@@ -92,12 +92,13 @@ class Catalog:
     def get_metric(self, metric_id: str) -> Metric:
         logging.debug(f"Looking up metric {metric_id!r} in catalog {self.id!r}")
         if metric_id not in self.metrics:
+            logging.info(f"[{metric_id}] Metric not found in catalog '{self.id}' — skipping")
```
tbittar, line 95: Why this log as we raise the error just the line after ?

gems_views_builder/input/input_data.py
```diff
@@ -0,0 +1,32 @@
...
+from gems_views_builder.input.view_config import ViewConfig
+
+
+@dataclass
+class InputData:
+    input_data_path: Path
+    taxonomy: Taxonomy
+    view_config: ViewConfig
+    catalogs: dict[str, Catalog]
+    library: Library
```
tbittar, line 30: We should be able to handle several libraries ? @aoustry I do not think this is already in one of the ADRs (and was not initially in the spec as needed input data). In fact we only need this to get the taxonomy category of each model of the lib

gems_views_builder/input/library.py
```diff
@@ -0,0 +1,79 @@
...
+            raise ValueError(f"Model {model_id} not found in library")
+
+    def get_taxonomy_category(self, model_id: str) -> str:
+        """Return the taxonomy category for a given model id."""
+        model = self.get_model(model_id)
+        if model.taxonomy_category is None:
+            raise ValueError(f"Model {model_id} has no taxonomy category in library")
+        return cast(str, model.taxonomy_category)
+
+    def get_components_in_taxonomy_category(self, taxonomy_category: str) -> list[str]:
```
tbittar:
```suggestion
    def get_models_in_taxonomy_category(self, taxonomy_category: str) -> list[str]:
```

gems_views_builder/input/library.py
```diff
@@ -0,0 +1,79 @@
...
+        port_types=parsed.port_types,
+        models={m.id: m for m in parsed.models},
+        models_by_taxonomy_category={
+            cat: [m.id for m in parsed.models if m.taxonomy_category == cat]
+            for cat in {m.taxonomy_category for m in parsed.models if m.taxonomy_category}
+        },
+    )
+
+
+def load_library_file(library_file_path: Path) -> LibrarySchema:
```
tbittar, line 69: Can't you already use parse_yaml_library of resolve_library from GemsPy ?

gems_views_builder/input/input_data.py
dusanparipovic: I catch the point and I agree. 
I suggest to add this after merge of all this PRs, it's already messy but I'll open issue. 
Then idea will be to to integrate `resolve_system` into `GVB`

gems_views_builder/input/catalog.py
dusanparipovic: I think this is better for consistency to have it also in log file.
If you don't agree we could remove

gems_views_builder/input/input_data.py
dusanparipovic: [Issue ](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/69)

gems_views_builder/views_builder.py
```diff
@@ -0,0 +1,46 @@
...
+
+from gems_views_builder.aggregator import Aggregator
+from gems_views_builder.input.input_data import InputData
+from gems_views_builder.metric_view import MetricView
+from gems_views_builder.metrics_builder import build_metric_structure
+from gems_views_builder.writer import MergedView, Writer
+
+
+class ViewBuilder:
+    def __init__(self, input_data: InputData) -> None:
```
aoustry, line 25: @dusanparipovic @guilpier-code 

You should have in mind that in a Study Folder, there might be several ViewConfig files (since a user might want to export several Views). 

- I guess we'll need a `view-config/` folder in the directory structure
- Shall we have several ViewConfig in Input data ? or should we just call the ViewBuilder several times with different arguments?

dusanparipovic: Yes right I have that on mind as Thomas said [here](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/62#issuecomment-4706705671) 
This is small adaptation but user needs to pass multiple view configs inside input folder. Then we could iterate over it and build separated views. 
Definitely we could proceed with simplification of folder structure, now we're strictly limited for input data structure.
We should have flexibility to say: simulation table is at path `fake_path/1/2/3` , taxonomy, etc...

gems_views_builder/metric_view.py
```diff
@@ -0,0 +1,26 @@
...
+# This file is part of the Antares project.
+
+import logging
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass
+class MetricView:
+    """Temporal aggregation result for a single metric, stored as a parquet file."""
```
aoustry:
```suggestion
    """View for a single computed metric, stored as a parquet file."""
```

gems_views_builder/time_aggregator.py
```diff
@@ -0,0 +1,71 @@
...
+    def run(self, metric_view: MetricView, metric: Metric) -> MetricView:
+        """
+        Step 2.C from POC[temporal aggregation]: Group by metric_id, metric_location, breakdown_properties, absolute_time_index, scenario
+        """
+        logging.info(f"[{metric.id}] Aggregating temporally with operator {metric.time_operator.value}")
+        lazy_metric_view = pl.scan_parquet(metric_view.file_path)
+        time_agg = (
+            pl.col("granular_metric_value").sum()
+            if metric.time_operator == TimeOperator.SUM
+            else pl.col("granular_metric_value").mean()
```
aoustry: Use elif, and add an else that raises an error if `metric.time_operator` has another value

gems_views_builder/input/input_data.py
```diff
@@ -0,0 +1,35 @@
...
+from gems_views_builder.input.system import System
+from gems_views_builder.input.taxonomy import Taxonomy
+from gems_views_builder.input.view_config import ViewConfig
+
+
+@dataclass
+class InputData:
+    input_data_path: Path
+    taxonomy: Taxonomy
+    view_config: ViewConfig
```
aoustry, line 28: Please open an Issue to think of having multiple ViewConfig in the future

gems_views_builder/common.py
```diff
@@ -1,12 +1,38 @@
...
 PARQUET_ROW_GROUP_SIZE = 64_000
 
+METRIC_STRUCTURE_TABLE_SCHEMA = pl.Schema(
+    {
+        "metric_id": pl.Utf8,
+        "component": pl.Utf8,
+        "metric_location": pl.Utf8,
+        "breakdown_properties": pl.Utf8,
+        "output": pl.Utf8,
+        "weight_output_id": pl.Int64,  # # What will be value range here, probably we could use Int8 to save memory?
```
aoustry:
```suggestion
        "weight_output_id": pl.Utf8,
```

resources/test_inputs/test_3/system.yml
aoustry, line 1: why nodes ?

aoustry (changes requested):

resources/test_inputs/test_3/system.yml
dusanparipovic: This is legacy nodes, tbh I don't know for what is used in past but new version of `gemspy` removed it, so pydantic will raise an error.

gems_views_builder/input/input_data.py
dusanparipovic: [Issue](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/71)

resources/test_inputs/test_3/system.yml
aoustry: Yes I guess we deleted this from GemsPy, you can now delete this.

gems_views_builder/views_builder.py
```diff
@@ -0,0 +1,55 @@
...
+                    logging.warning(f"Metric {metric_id} not found in catalog {catalog_id}")
+                    continue
+
+                metric_structure_table = self.metric_structure_table_builder.build(metric)
+                metric_view = self.terms_aggregator.run(metric_structure_table, metric)
+                temporal_metric_view = self.time_aggregator.run(metric_view, metric)
+                metric_views.append(temporal_metric_view)
+
+        # # Final step merging all metric views into one
+        return MergedView(self.input_data.results_path, metric_views)
```
guilpier-code:
```python
return MergedView(self.input_data.results_path, metric_views)
```
- Here we exhibit a path` self.input_data.results_path`, which is a detail we don't want to know about at this level (this a technical consideration, not a business consideration).
  So **what can we do about this** ?
  Printing on disk the result of an algorithm and printing on disk for persistence is not the same thing. We want to hide persistence as much as possible, but at some point, we want to show the results and that we save them on disk (not in the building views algorithm, though). We are currently treating the printing of the results as if it was a persistence consideration. Its' not.
  So I would recommend that we `return MergedView(metric_views)`
  We could return a object **MergedView** that just stores a list of **MetricView** or you can also return a list of **MetricView**. My preference goes to the first solution, but it's as you like.
  Then we can print it, outside the algorithm (see further)   
- There is another strange thing : we store results (or output) path inside the **input_data**. They're supposed to be incompatible : input **and** output.

In the end, we could have (in a main program for instance) : 
```python
input_data = ... # building the input
output_path = ... # getting this from somewhere (a **CLI** option maybe)
viewBuilder = ViewBuilder(input_data)
mergedView = viewBuilder.build()
Save(mergedView, output_path)
```
Note that : 
- we don't mix saving (technical) and merged view (business) : class **MergedView** does not have a save method
- **output_path** could be retrieved from the CLI options

gems_views_builder/merged_view.py
```diff
@@ -0,0 +1,37 @@
...
+import polars as pl
+
+from gems_views_builder.common import PARQUET_COMPRESSION, PARQUET_COMPRESSION_LEVEL, PARQUET_ROW_GROUP_SIZE
+from gems_views_builder.metric_view import MetricView
+
+
+class MergedView:
+    """Final merged view written to the results directory."""
+
+    def __init__(self, result_path: Path, metric_views: list[MetricView]) -> None:
```
guilpier-code: On class **MergedView**, please see [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/66#discussion_r3481573144)

gems_views_builder/input/input_data.py
```diff
@@ -0,0 +1,35 @@
...
+    catalogs: dict[str, Catalog]
+    library: Library
+    system: System
+    filtered_st: FilteredSimulationTable
+    results_path: Path
```
guilpier-code: We shouldn't mix input data and result path (=output).
On that matter, see [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/66#discussion_r3481573144)

gems_views_builder/metric_view.py
```diff
@@ -0,0 +1,26 @@
...
+import logging
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass
+class MetricView:
+    """View for a single computed metric, stored as a parquet file."""
+
+    file_path: Path
```
guilpier-code:
```python
file_path: Path
```
prefer **persistence_path**, more accurate.

gems_views_builder/views_builder.py
aoustry: Do we really need a "MergedView" class? I mean, the targeted terminology for the final product of this workflow is a "View", not a "MergedView"... This term was built up during development! I guess it would be cleaner to get rid of it...

@dusanparipovic @guilpier-code

dusanparipovic: We could remove it and then main algorithm will return list of MetricView and we will call `save` function directly on it that's what I'm considering now. 
Now purpose of `MergedView` is only container for list of `MetricViews`

dusanparipovic: What do you think @guilpier-code, also I could then remove save function and delete `merged_view.py`, I could put it inside common.py ?

gems_views_builder/common.py
```diff
@@ -1,23 +1,72 @@
...
+        "output": pl.Utf8,
+        "weight_output_id": pl.Utf8,
+    }
 )
+
+
+LOG_DIR = Path(__file__).resolve().parent.parent / "logs"
+
+
+def configure_logging(verbose: bool = False, log_dir: Path | None = None) -> Path:
```
guilpier-code: Sorry, but it seems that the last **merge** commit went wrong : 
here we end up with the a way to configure the logging that we chose to replace with a simpler one in a previous PR.

gems_views_builder/views_builder.py
```diff
@@ -0,0 +1,54 @@
...
+                except ValueError:
+                    logging.warning(f"Metric {metric_id} not found in catalog {catalog_id}")
+                    continue
+
+                metric_structure_table = self.metric_structure_table_builder.build(metric)
+                metric_view = self.terms_aggregator.run(metric_structure_table, metric)
+                temporal_metric_view = self.time_aggregator.run(metric_view, metric)
+                metric_views.append(temporal_metric_view)
+
+        # # Final step merging all metric views into one
```
guilpier-code:
```python
 # # Final step merging all metric views into one
```
This comment is out of date. We should just remove it.

gems_views_builder/views_builder.py
guilpier-code: What @dusanparipovic did at this stage seems ok to me.
To enhance a bit, we could rename the **save** function into **accumulate_on_disk**, more accurate.
(we don't really "merge", we "append" or "accumulate")

tests/test_filter_simulation_table.py
dusanparipovic: Agree. 
I asked Antoine regarding this, I'll open GitHub issue but this isn't highest priority now.

gems_views_builder/common.py
dusanparipovic: Yes right I didn't think about CLI at that time, sorry for that in fact currently we also use this function inside CLI.

guilpier-code: ok, if you think this logging configuration is more suitable.
It wasn't reviewed though, here is my (short) review : 
- `root` is a logger, so we should renamed it into `logger`
- We `return log_file` but we don't use it. So : 
  - either we just remove it
  - or we build this file path outside the logging configuration (cleaner) : 
 ```python
    args = build_parser().parse_args(argv)
    log_file = make_log_file(log_dir=args.log_dir)
    configure_logging(verbose=args.verbose, log_file)
```
Doing this, **configure_logging(...)** is shorter and focuses on builtin module **logging** settings (handlers, formatters, level).

guilpier-code (approved):

aoustry (approved):

PR #70, opened 2026-06-25
feat(CLI): introduce command line interface

### LINKED ISSUE #67: feat(CLI): introduce command line interface

dusanparipovic:


aoustry (approved):

gems_views_builder/cli.py
guilpier-code, line 1: Something is strange here : 
We are in the **cli.py** source file.
So we could expect from this file to contain only **CLI** things.
As a consequence, only leave `build_parser()` should and the checks on parsed options should remain in this file.
The rest should go to the main program source file (**\_\_main\_\_.py** ?)

gems_views_builder/__main__.py
guilpier-code, line 1: Following [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/70/changes#r3480644804) (from **cli.py**'s review), the main program should call : 
- the parsing of options (CLI) from **cli.py**
- the checking of options from **cli.py**
- configuring the logging
- function **run** (this function could be defined in the **\_\_main\_\_.py** or in a **run.py**)

gems_views_builder/cli.py
dusanparipovic: Very nice catch, thank you.
Done

gems_views_builder/common.py
```diff
@@ -33,17 +34,25 @@
 )
 
 
-log_file = (
```
guilpier-code, line 36: I guess that what will remain from the logging configuration in the end is the `logging.basicConfig(...)`, and not what's in use in the **common.py** here

guilpier-code (approved):

PR #72, opened 2026-06-30
fix(temporal aggregation): add date truncate to temporal aggregation

### LINKED ISSUE #59: bug: Temporal aggregation is ignored since no date truncation applied
## Step 2c — Temporal aggregation

Source: `aggregator.py:Aggregator.aggregate_metric_temporally`

```sql
SELECT
  metric_id, metric_location, breakdown_properties, scenario,
  granular_date AS view_date,
  TIME_OPERATOR(granular_metric_value) AS metric_value
FROM metric_view
GROUP BY metric_id, metric_location, breakdown_properties, scenario, view_date
```

`TIME_OPERATOR` = `SUM` or `AVG` from `metric.time_operator`.

## **Current limitation**: `view_date` = `granular_date` (no date truncation applied).
The `time` field in `view_config.aggregation` (`hour|day|week|month|year`) is loaded but not yet passed to the aggregator. See [ADR-007](../adr/007-consistency-check-strategy.md).

dusanparipovic:


gems_views_builder/time_aggregator.py
```diff
@@ -73,3 +80,19 @@ def run(self, metric_view: MetricView, metric: Metric) -> MetricView:
         )
         logging.info(f"[{metric.id}] Temporal aggregation written to {out_path}")
         return MetricView(out_path)
+
+    @staticmethod
+    def parse_time_aggregation(time_aggregation: TimeAggregation | None) -> str:
```
aoustry: What is the expected input syntax ?

aoustry: (as regards the str for viewconfig.time_aggregation)

tests/test_time_aggregator.py
```diff
@@ -44,28 +47,68 @@ def _metric(time_operator: TimeOperator) -> Metric:
...
     assert first.persistence_path != second.persistence_path
     assert first.persistence_path.name.endswith("-0.parquet")
     assert second.persistence_path.name.endswith("-1.parquet")
+
+    # Each part is truncated to its day and the two intra-day values are summed.
+    for part in (first, second):
+        df = pl.read_parquet(part.persistence_path)
+        assert df.shape[0] == 1
+        assert df["view_date"][0] == datetime(2026, 1, 1, 0, 0)
+        assert df["metric_value"][0] == approx(30.0)
```
aoustry, line 166: One should add a e2e test that returns a view resulting from temporal aggregation (ex: Daily granularity instead of hourly) and check that the numerical values matches (as regards load or production for instance).

aoustry (changes requested):

tests/test_time_aggregator.py
aoustry: One should also check with a metric, the aggregation operator of which is the average for the time dimension (ex: prices).

gems_views_builder/time_aggregator.py
aoustry: Suggested section in the ViewConfig files (YAML).

```
  aggregation:
    - time: hour | day | week | month | year
  ```

Comment (dusanparipovic) [reply gems_views_builder/time_aggregator.py]: > What is the expected input syntax ?

```yml
view:

  id: view_area

  scope:
    - location:
      taxonomy-category: balance

    - calendar: calendar_file 

  aggregation:
    - time: hour

  catalog: #Pointer to "catalogs" = "Business Metrics definitions files".
      - id: catalog

  metrics:  
      - id: catalog.PROD
      - id: catalog.LOAD
      - id: catalog.BALANCE
```

dusanparipovic: To be honest this syntax from POC isn't that intuitive for me, we could change it to be something like: 
Instead of 
```yaml
  aggregation:
    - time: hour
```
Proposition[1]: 

```yaml
  time-resolution: day | week | month | year
```
Proposition[2]:
```yaml
  view-date-resolution: day | week | month | year
```
What do you think @aoustry @guilpier-code @tbittar ?

aoustry: I agree that my proposal can be improved. However, please keep mind that the viewconfig file should NOT be too flat -> we should pay attention to structure it. 
In my proposal, we have the following sections

- **scope**, that is supposed to describe the spatial/temporal/scenario domain -> the subsection are location, calendar and soon scenario.
- **aggregation**, that is supposed to describe the aggregation for the spatial/temporal/scenario domains -> the subsection could be spatial, time scenario.

We might structure the file the other way around (sections could be spatial/time/scenario, and subsection could be scope/aggregation)... but anyway, we have to avoid a flat list of parameters I guess.

dusanparipovic: Right, I agree with you. 
Then we definitely need to follow your structure, it's more structured. 
I would like only to change `time` -> `time-resolution`/`view-date-resolution`it's more intuitive in my opinion for end user. 
Thank you for clarification!!!!!

gems_views_builder/time_aggregator.py
```diff
@@ -8,11 +8,22 @@
...
+    TimeAggregation.DAY: "1d",
+    TimeAggregation.WEEK: "1w",
+    TimeAggregation.MONTH: "1mo",
+    TimeAggregation.YEAR: "1y",
+}
+
 
 class TimeAggregator:
-    def __init__(self) -> None:
+    def __init__(self, time_aggregation: TimeAggregation | None) -> None:
```
guilpier-code, line 25: If I understand you correctly, we could have : 
```python
def granularDateColumn(time_aggregation: TimeAggregation | None)
    if time_aggregation:
        try:
            granular_date.dt.truncate(TRUNCATE_WINDOWS[time_aggregation]).alias("view_date")
        except KeyError:
            raise ValueError(...)
    else:
        pl.col("granular_date").alias("view_date")
...

class TimeAggregator:
    ...    
    def run(self, metric_view: MetricView, metric: Metric) -> MetricView:
        lazy_metric_view = ...
        aggregation_op = timeAggregationOperator(metric) # Free function
        granular_date_col = granularDateColumn(self._time_aggregation)
        view = (
              lazy_metric_view.with_columns(granular_date_col)
              .group_by(...)
              .agg(aggregation_op)
              .select(...)
        ) 
       ....
```
am I wrong ?
If I'm right, it would be a bit more clear (we hide details from **run(...)** function).

gems_views_builder/time_aggregator.py
guilpier-code, line 36: `atexit.register(rmtree, self._root_dir, True)` : 
Good ! 
I didn't know this was possible. Handy.

guilpier-code, line 76: `self._part_counter += 1` : 
I must conclude that` metric.id` isn't enough to make that saving unique, right ?
We need a counter then ?

gems_views_builder/time_aggregator.py
```diff
@@ -73,3 +89,13 @@ def run(self, metric_view: MetricView, metric: Metric) -> MetricView:
         )
         logging.info(f"[{metric.id}] Temporal aggregation written to {out_path}")
         return MetricView(out_path)
+
+    @staticmethod
```
guilpier-code: Regardless of comments above, this **parse_time_aggregation(...)** should be a free function : it does not use any data member

tests/test_time_aggregator.py
```diff
@@ -44,28 +47,72 @@ def _metric(time_operator: TimeOperator) -> Metric:
...
+        (None, None),
+    ],
+)
+def test_parse_time_aggregation(aggregation: TimeAggregation | None, expected_window: str) -> None:
+    assert TimeAggregator.parse_time_aggregation(aggregation) == expected_window
+
+
+def test_parse_time_aggregation_invalid_raises() -> None:
+    with pytest.raises(ValueError, match="Invalid time aggregation"):
+        TimeAggregator.parse_time_aggregation("decade")  # type: ignore[arg-type]
```
guilpier-code: If we apply comments from **time_aggregator.py**'s review, can we test free function **granularDateColumn(...)** easily (instead of **parse_time_aggregation(...)**) ?

tests/test_time_aggregator.py
dusanparipovic: To be clear [issue](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/76)
This is in general tests refactoring + your proposal

gems_views_builder/time_aggregator.py
dusanparipovic: Right but to be more precise `aggregator operator` and `granular date column` will be polars expression, so I would prefer names `time_aggreation_expression` and `granular_date_expression`
What do you think about names, does it fit for you ? 
But your idea in general is very good.

tests/test_time_aggregator.py
dusanparipovic: Right, also added test case for each possbile situation situation.

gems_views_builder/time_aggregator.py
dusanparipovic: Why this could be problematic? 
Imagine we have 2 metrics in 2 different catalogs with same name, but different term definition. 
For example
`Catalog A`
```yaml
  - id: PROD
    terms:
      - taxonomy-category: production
        output-id: p
        location-ports: p_balance_port
    terms-operator: sum
    time-operator: sum
```
`Catalog B`
```yaml   
  - id: PROD
    terms:
      - taxonomy-category: production / whatever
        output-id: p /whatever
        location-ports: p_balance_port / whatever
    terms-operator: avg / whatever 
    time-operator: avg / whatever 
```
Metric from Catalog A and Catalog B could be completely same but without part counter second processed metric will be overwritten. 
This is prevention that will give user ability to have `"namespace"` inside catalog.
Potentially what could be better solution is to have  check of metric reference name in view config file, with this we will lose need for `part_counter` and user will lose ability to have same metric ID in 2 different catalogs.

gems_views_builder/time_aggregator.py
guilpier-code: ok, I agree with that

gems_views_builder/time_aggregator.py
guilpier-code: ok, no problem with that, at least for now. 
In the end we accumulate these **time aggregated views** without the need of identifying which parquet file is related to which **metric view**.
Maybe some day there will be a new treatment on those tables that requires to be more precise. But this one hasn't come yet.

gems_views_builder/time_aggregator.py
```diff
@@ -73,3 +82,21 @@ def run(self, metric_view: MetricView, metric: Metric) -> MetricView:
         )
         logging.info(f"[{metric.id}] Temporal aggregation written to {out_path}")
         return MetricView(out_path)
+
+
+def granular_date_expression(time_aggregation: TimeAggregation | None) -> pl.Expr:
```
guilpier-code, line 87: `def granular_date_expression(...)` : 
Here we raise an exception in case **time_aggregation**'s value is incorrect.
It raises one question for me : this correctness could have been checked earlier (for example after reading **view_config.yml**), isn't it ?
If what I'm saying is true, this check does have to be made here.
It implies a simplification : 
```python
def granular_date_expression(time_aggregation: TimeAggregation | None) -> pl.Expr:
    if time_aggregation:
        return pl.col("granular_date").dt.truncate(TRUNCATE_WINDOWS[time_aggregation]).alias("view_date")
    return pl.col("granular_date").alias("view_date")

```

dusanparipovic: If user provide bad value in view config file, pydantic will automatically drop error if value isn't in this set {"hour", "day", "mont", "year"}
If value isn't provided that's fine pydantic it will treat as None and we won't apply truncation. 
When we come to this function we know that we have valid input which is:

- `None`  - nothing writtent in view config, so we will assume not truncation
- `TimeAggregation`- perform correct truncation

dusanparipovic: What this mean practically this function won't come in situation to raise any error.

tests/test_time_aggregator.py
```diff
@@ -10,30 +10,48 @@
...
+    df = pl.DataFrame({"granular_date": [date]}, schema={"granular_date": pl.Datetime})
+    return cast(datetime, df.select(granular_date_expression(aggregation)).item())
+
+
+def _apply_agg_expr(values: list[float], time_operator: TimeOperator) -> float:
+    df = pl.DataFrame({"granular_metric_value": values})
+    return float(df.select(time_aggregation_expression(time_operator)).item())
+
+
+def _granular_view(rows: list[tuple[datetime, float]], tmp_path: Path) -> MetricView:
```
guilpier-code: `def _granular_view` 
this function builds a **MetricViews**, so it's a kind of factory.
we could rename it **makeMetricView** or **createMetricView**, what do you think ?

tests/test_time_aggregator.py
guilpier-code: `def _metric` is a metric factory (it makes a **Metric**).
It could be renamed into **makeMetric** or **createMetric**.

tests/test_time_aggregator.py
```diff
@@ -10,30 +10,48 @@
...
 
 from gems_views_builder.input.catalog import Metric, TermsOperator, TimeOperator
+from gems_views_builder.input.view_config import TimeAggregation
 from gems_views_builder.metric_view import MetricView
-from gems_views_builder.time_aggregator import TimeAggregator
+from gems_views_builder.time_aggregator import (
+    TimeAggregator,
+    granular_date_expression,
+    time_aggregation_expression,
+)
```
guilpier-code, line 30: I see that **"_"** are prepended.
It's usually for private methods.
In our case, what is it for ?

tests/test_time_aggregator.py
```diff
@@ -44,28 +62,85 @@ def _metric(time_operator: TimeOperator) -> Metric:
...
+    [
+        (TimeOperator.SUM, [10.0, 20.0], 30.0),
+        (TimeOperator.AVG, [10.0, 20.0], 15.0),
+    ],
+)
+def test_time_aggregation_expression(time_operator: TimeOperator, values: list[float], expected: float) -> None:
+    assert _apply_agg_expr(values, time_operator) == approx(expected)
+
+
+def test_truncation_groups_by_window(tmp_path: Path) -> None:
```
guilpier-code, line 94: If we wanted to be a bit more clear, we could separate the 3 steps of a unit test, the 3 **A**'s : 
- Arrange
- Act
- Assert

For example (3 A's in comments are not mandatory, but welcome) : 
```python
def test_truncation_groups_by_window(tmp_path: Path) -> None:
    # Arrange
    aggregator = TimeAggregator(TimeAggregation.DAY)
    rows = [(datetime(2026, 1, 1, 3, 0), 10.0), (datetime(2026, 1, 1, 20, 0), 20.0)]
    metricView = makeMetricView(rows, tmp_path)
    metric = makeMetric(TimeOperator.SUM)
    
    # Act
    outMetricView = aggregator.run(metricView, metric)
   
    # Assert 
    df = pl.read_parquet(outMetricView .persistence_path)
    assert df.shape[0] == 1
    assert df["view_date"][0] == datetime(2026, 1, 1, 0, 0)
    assert df["metric_value"][0] == approx(30.0)
    assert df["metric_value"].dtype == pl.Float64
```

tests/test_time_aggregator.py
dusanparipovic: Yes right, there's no reason for _

tests/test_time_aggregator.py
dusanparipovic: Right agree, I'll follow this in future.
Thank you for remark.

aoustry (changes requested):
We would need a e2e test where we parse a YAML where the value of time-aggregation is parsed and effectively leads to an aggregation.

aoustry (approved):

guilpier-code (approved):

PR #77, opened 2026-07-03
feat(simulation table csv): add support for csv simulation table 

dusanparipovic:


gems_views_builder/view.py
```diff
@@ -14,27 +14,43 @@
...
-    result_path = results_path / f"view{timestamp}.parquet"
-    pl.scan_parquet([v.persistence_path for v in metric_views]).sink_parquet(
-        result_path,
-        compression=PARQUET_COMPRESSION,
-        compression_level=PARQUET_COMPRESSION_LEVEL,
-        row_group_size=PARQUET_ROW_GROUP_SIZE,
-    )
+    merged = pl.scan_parquet([v.persistence_path for v in metric_views])
+
+    if output_format == "parquet":
```
guilpier-code: `output_format`'s value (**parquet** or **csv**)  should be checked when checking CLI options, not when printing.
Indeed : we made the whole view building, now we want to print... and exception is raised. No view printing.
Maybe we would have known about the wrong format earlier.

gems_views_builder/__main__.py
guilpier-code, line 47: _**Just an idea : adopt it if you like it (overkill ?).**_

following [this earlier comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/77/changes#r3521002314), we could even : 
- Build a very light view sinker here (different depending on we print into **parquet** or **csv**, but with the same interface)
- pass it to **run** function and then to **accumulate_on_disk**

The benefits are : 
- **run** takes less arguments
-  **accumulate_on_disk** takes less arguments and gets smaller. It does not know in which format it prints, or even where it prints (location on disk) : it just merges the views, sinks it, then scans into a data frame it and makes it a View to return. But it does not know about format and destination path for sinking. If we change the format or destination  **accumulate_on_disk** remains unchanged.
- we have split code in many simpler pieces

```python
# in main program
def main(argv: list[str] | None = None) -> int:
    args = build_parser().parse_args(argv)
    configure_logging(verbose=args.verbose, log_dir=args.log_dir)
    error = check_options(args)
    ...
    view_sinker = viewSinkerFactory(args.results_dir, args.output_format).make()
    ...
    run(args.input_dir, view_sinker)
    ...

# in view.py
def accumulate_on_disk(metric_views: list[MetricView], viewSinker: ViewSinker) -> View:
    timestamp = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%S")
    merged = pl.scan_parquet([v.persistence_path for v in metric_views])
    viewSinker.sink(merged)
    dataframe = pl.scan_parquet(viewSinker.destination())
    return View(dataframe=dataframe)
```

aoustry (approved):

gems_views_builder/__main__.py
dusanparipovic: Done! 
Additionally I created one new folder `view` where we can store all view related things, what do you think about it ?

gems_views_builder/cli.py
```diff
@@ -57,4 +65,7 @@ def check_options(args: argparse.Namespace) -> int | None:
     if not args.results_dir.is_dir():
         logging.error(f"Results directory does not exist: {args.results_dir}")
         return 2
+    if args.output_format not in {"parquet", "csv"}:
```
guilpier-code: **Question** :
here we check value of `args.output_format`.
But parser is given choices : 
```python
parser.add_argument(
       ...
       choices=["parquet", "csv"],
       ...
    )
```
So my question : is it not redundant ?

tests/test_filtering_and_breakdown.py
```diff
@@ -46,7 +46,8 @@ def filtering_and_breakdown_workspace(test_files_root: Path, tmp_path: Path) ->
     results_dir.mkdir()
     shutil.copytree(src, dst)
     metric_views = ViewBuilder(Loader(dst).load()).build()
-    view = accumulate_on_disk(metric_views, results_dir)
+    sinker = ViewSinkerFactory(results_dir, "parquet").make()
```
guilpier-code:
```python
sinker = ViewSinkerFactory(results_dir, "parquet").make()
```
We don't need the factory here, we should have instead : 
```python
sinker = ParquetViewSinker(results_dir)
```

gems_views_builder/view/view_sinker.py
```diff
@@ -0,0 +1,52 @@
...
+from pathlib import Path
+
+import polars as pl
+
+from gems_views_builder.common import PARQUET_COMPRESSION, PARQUET_COMPRESSION_LEVEL, PARQUET_ROW_GROUP_SIZE
+from gems_views_builder.view.view import View
+
+
+class ViewSinker(ABC):
+    def __init__(self, output_path: Path, output_format: str):
```
guilpier-code: `def __init__(self, output_path: Path, output_format: str):`

We should remove **output_format** : 
- from **ViewSinker**'s constructor arguments list
- also as a **ViewSinker** data member.

Indeed, each sinker (**csv** or **parquet**) is supposed to know the output format it prints files into.
For example : 
```python
class ParquetViewSinker(ViewSinker):
    def sink(self, merged: pl.LazyFrame) -> View:
        result_path = self.output_path / f"view{self.timestamp}.parquet"
        merged.sink_parquet(
            ...
        )
        logging.info(f"Results merged into parquet file")
        return View(dataframe=pl.scan_parquet(result_path))
```

tests/test_filtering_and_breakdown_property_order.py
```diff
@@ -47,7 +47,8 @@ def property_order_workspace(test_files_root: Path, tmp_path: Path) -> tuple[Pat
     results_dir.mkdir()
     shutil.copytree(src, dst)
     metric_views = ViewBuilder(Loader(dst).load()).build()
-    view = accumulate_on_disk(metric_views, results_dir)
+    sinker = ViewSinkerFactory(results_dir, "parquet").make()
```
guilpier-code: Same remark as for **test_filtering_and_breakdown.py**'s review

tests/test_view_builder.py
```diff
@@ -40,7 +39,8 @@ def view_result(test_files_root: Path, tmp_path: Path) -> pl.DataFrame:
     results_dir.mkdir()
     shutil.copytree(src, dst)
     metric_views = _build_view_builder(dst).build()
-    accumulate_on_disk(metric_views, results_dir)
+    sinker = ViewSinkerFactory(results_dir, "parquet").make()
```
guilpier-code: `sinker = ViewSinkerFactory(results_dir, "parquet").make()` : create a **ParquetViewSinker** directly instead

tests/test_view_builder.py
```diff
@@ -120,6 +120,29 @@ def test_balance_busb_values(view_result: pl.DataFrame) -> None:
     assert rows["metric_value"].to_list() == expected
 
 
+# ---------------------------------------------------------------------------
+# Output format
+# ---------------------------------------------------------------------------
+
+
+def test_accumulate_on_disk_writes_csv(test_files_root: Path, tmp_path: Path) -> None:
```
guilpier-code: Here I would : 
- highlight important of a test (the 3 **A**'s : arrange, act, assert)
- highlight important steps in each test step
- hide details

```python
def build_view_from_parquet_simu_table___print_it_in_csv___check_view_format(test_files_root: Path, tmp_path: Path) -> None:
   # Arrange 
    src = test_files_root / "test_3"
    dst = copy_study_in_a_tmp(src)
    sinker = CsvViewSinker(results_dir)
    
    # Act
    run(dst, sinker) # form the main program
    
    # Assert
    result_files = list(dst.glob("view*.csv"))
    assert result_files, "No result csv file written"
    assert not list(dst .glob("view*.parquet")), "Unexpected parquet file written"
    assert pl.read_csv(result_files[0]).height > 0
```
Note : 
- we don't need to have 2 output path (dst + results_dir)
- are the results of test removed ? Otherwise, they should be : a way to do that is to use a fixture with a tear down, that deletes **dst**. 
- Here I use **run(...)** (which could have a better name) from the main program.

tests/test_view_builder.py
```diff
@@ -120,6 +120,29 @@ def test_balance_busb_values(view_result: pl.DataFrame) -> None:
     assert rows["metric_value"].to_list() == expected
 
 
+# ---------------------------------------------------------------------------
+# Output format
+# ---------------------------------------------------------------------------
+
+
```
guilpier-code: We don't need comments in tests (and in any code unit) if code is clear enough, starting by test title.
```python
def build_view_from_parquet_simu_table___print_it_in_csv___check_view_format(test_files_root: Path, tmp_path: Path) -> None:
   ...
```

tests/test_view_builder.py
```diff
@@ -136,7 +159,8 @@ def test_log_messages_emitted_to_stdout(
     results_dir.mkdir()
     with caplog.at_level(logging.INFO):
         metric_views = _build_view_builder(dst).build()
-        accumulate_on_disk(metric_views, results_dir)
+        sinker = ViewSinkerFactory(results_dir, "parquet").make()
```
guilpier-code: Create directly a view sinker.

gems_views_builder/view/view_sinker.py
dusanparipovic: Right sorry, for that miss, that's whole point.

gems_views_builder/cli.py
dusanparipovic: Yes this isn't needed check, it will raise and error.
I forgot to delete this line for `check_arguments` while adopting.

tests/test_view_builder.py
dusanparipovic: Pytest will delete automatically everything what lives under `tmp_path`
What isn't cleaned is `pipeline log` but it's already in `git ignore`. 
For output now everything is stored in one path.

guilpier-code (approved):

PR #78, opened 2026-07-06
refactor(metric structure builder): simplification of workflow and metric structure table building

dusanparipovic:
Still in progress, tiny additional improvements needed


gems_views_builder/input/system.py
```diff
@@ -18,7 +18,7 @@
 from typing import Any, cast
 
 from gems.model.library import Library as GemsLibrary  # type: ignore
-from gems.study import Component  # type: ignore
+from gems.study import Component as GemsComponent  # type: ignore
```
guilpier-code: **GemsComponent** ==> **GemsPyComponent** : more accurate

same for **System**

gems_views_builder/metrics_structure_builder.py
```diff
@@ -74,45 +50,29 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
-                                "output": term.output_id,
-                                "weight_output_id": 1,
-                            }
-                        )
-                    else:
-                        logging.debug(
-                            f"[{metric.id}] Component {component_id!r} did not match metric filter and was skipped"
-                        )
 
+            for component_id, component in self.components_by_taxonomy_category[term.taxonomy_category].items():
```
guilpier-code: I'm sorry I wasn't clear enough in the [core comment in PR 51](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/51#discussion_r3511504671)
So please let me try to be more clear. 
The core of my opinion is : we may not need **components_by_taxonomy_category** at all.
The pieces of information contained in this dictionary are already (or should be) contained by each component : each component inherits its **taxonomy category** from its model.
Each component knows (or should know) if it matches a given **taxonomy category**. It also knows if it matches a metric filter.
If I rephrase : **components_by_taxonomy_category** is external to components, but the piece of information it contains should be held by components themselves, because it's consistent and natural. Every component has a **taxonomy category** inherited from its model.
Except if I miss something of course, on that matter, **specifications** (by forcing to define a function **LIST_OF_COMPONENTS_IN_TAXONOMY_CATEGORY**) may not be optimal.

So, instead of having : 
```c++
    def build(self, metric: Metric) -> MetricStructureTable:
        ...
        for term in metric.terms:
            for c in self.components_by_taxonomy_category[term.taxonomy_category].items():
                if c.check_component_filter_matches(metric.filter):
                    ...
```
We should have : 
```c++
    def build(self, metric: Metric) -> MetricStructureTable:
        ...
        for term in metric.terms:
            for c in self.components if c.match(term.taxonomy_category) and c.match(metric.filter) :
                    ...
```
In my opinion, the second code block is simpler and more natural.

gems_views_builder/metrics_structure_builder.py
```diff
@@ -74,45 +50,29 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
+                if component.check_component_filter_matches(metric.filter):
```
guilpier-code: **c.check_component_filter_matches(...)** : very long name.
We repeat **component** : **c** is a component, so we don't need the **component** in the method's name.
I would suggest : **c.matches(...)** (with an expected filter as argument)

gems_views_builder/__main__.py
```diff
@@ -29,10 +37,30 @@ def run(input_dir: Path, results_dir: Path) -> None:
...
+    # # Group components by taxonomy category
+    components_by_taxonomy_category = group_components_by_taxonomy_category(components)
+
+    # # Build component-port connection index
+    component_port_connections = build_component_port_connections(input_data.system.connections)
+    save_component_port_connections(components, component_port_connections)
+
+    # # Only one instance of MetricStructureTableBuilder is needed
+    metric_structure_table_builder = MetricStructureTableBuilder(
+        input_data.view_config.location_taxonomy_category,
```
guilpier-code, line 55: `input_data.view_config.location_taxonomy_category` : 
It be renamed into `input_data.view_config.taxonomy_category`, shorter

And you know what ? For now I can't tell why we have a **scope**.
A **view config** could have only a **calendar** and a **location**.  
Having a **scope** level may seem useless. But I may miss something.

gems_views_builder/metrics_structure_builder.py
```diff
@@ -74,45 +50,29 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
+                    component_contribution_locations = component.get_location(location_ports=term.location_ports)
```
guilpier-code: **We talked with @aoustry, we agreed on this :** 

Given a component, we build the list of locations associated with the components connected to this candidate component (via the port specified in the metric), those locations must match the location defined by the **view config** (not the one defined by the metric term).

Regarding this step, here's what currently happens (directly in the algorithm that generates a **METRIC_STRUCTURE_TABLE**):
- For a given component, and for the port referenced in the metric term, we iterate through all the connections of that component via this port.
- If any of the connected components does not belong to the taxonomy category defined by the **view config**, an exception is thrown (which terminates the program because the exception is not caught).

Personally, I'm not comfortable with this approach.

- Throwing an exception in the middle of a computation because we encounter an issue that wasn't anticipated beforehand isn't ideal. We could detect this kind of problem before starting entering the view building algo, after loading input data.
- After loading the input data (**Library** and **System**), we could iterate through the connections of every component (via its ports), and for each port determine a location that would be stored within the component itself (note that here we're being exhaustive by processing every port of every component—see the trade-off below).
- While doing so, we would verify that each port is associated with only one location (within the component). Otherwise, either we have an actual inconsistency (and throw an exception), or
no location can be determined for a given port. In that case, the algorithm would simply skip the component that references this port.
- Trade-off (as mentioned above): at the beginning of the computation, we would exhaustively iterate through the connections associated with every port of every component, instead of only iterating through the connections associated with the specific metric port (as we do today). So there is a performance cost. However, we gain several advantages:
  - potential issues are detected much earlier (for example, inconsistencies between the locations of connections associated with the same port);
  - the code for building the **METRIC_STRUCTURE_TABLE** becomes simpler.

---

Back to code. So we may have : 
```c++
    def build(self, metric: Metric) -> MetricStructureTable:
        ...
        for term in metric.terms:
            for c in self.components_by_location[term.taxonomy_category] if c.match(metric.filter) \
                                   and c.is_located_at(term.location_ports, self.view_cfg_location):
                rows.append(
                    {
                        "metric_id": metric.id,
                        "component": c.id(),
                        "metric_location": c.formatted_locations(term.location_ports, self.view_cfg_location),
                        "breakdown_properties": c.matching_breakdown(metric.breakdown),
                        "output": term.output_id,
                        "weight_output_id": 1,
                    }
               )
        return MetricStructureTable(rows, metric)
```
Note that :
- about `c.is_located_at(term.location_ports, self.view_cfg_location)` : a component should already contain a **dictionary** that associates to a pair (port_id, taxonomy_category) a unique component id (or location) our contributing (= emitting) component is connected to, via **port_id**. This dictionary could quickly answer the question : do I have a key matching the pair made of the 2 arguments of this method ?
- about `c.formatted_locations(term.location_ports, self.view_cfg_location)` : the same dictionary would allow to fetch the associated component ids  (= locations), one per location port, and format them as we want.

gems_views_builder/input/system.py
```diff
@@ -39,21 +39,19 @@ def __init__(self, system: GemsSystem) -> None:
...
 
     @property
     def connections(self) -> list[Any]:
         return cast(list[Any], getattr(self._system, "connections", None) or [])
 
+    # # This could be removed also
+    # # I would like to keep this function as reminder
+    # # When time comes we will be able to have multiple libraries as input so we need to keep distinction
+    # # between components which will have same id but they will be in different libraries with different definitions
     def _models_to_components(self) -> dict[str, list[str]]:
```
guilpier-code: **_models_to_components** : seems unused

gems_views_builder/metrics_structure_builder.py
```diff
@@ -74,45 +50,29 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
+                    self._location_components_match_taxonomy_category(component_contribution_locations)
+                    rows.append(
+                        {
+                            "metric_id": metric.id,
+                            "component": component.id,
+                            "metric_location": _format_metric_location(component_contribution_locations),
+                            "breakdown_properties": component._format_breakdown_properties(metric.breakdown),
```
guilpier-code: `component._format_breakdown_properties` : when I see a prepending underscore, I think : "it's private".
But here, it's not, it's called from outside, so it's **public**.

gems_views_builder/metrics_structure_builder.py
aoustry: @guilpier-code do you think it is equivalent in terms of performance?

dusanparipovic: Your approach is good and natural from perspective of reading as I said earlier. 
But in terms of algorithm this isn't performant in fact we will perform `O(N*M)` where N is number of components and M number of metrics. Instead of `O(n*M)` where **small n** is subset of components which belongs to taxonomy category. 
Real gems/hybrid study has 2000 components.

guilpier-code: ok, so let's keep the idea of grouping components be taxonomy categories, for performance reasons.
So I may update some other existing review comments in order to take this into account.
Especially [this one](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/78#discussion_r3538119334)

guilpier-code: With an update, we may have : 
```c++
    def build(self, metric: Metric) -> MetricStructureTable:
        ...
        for term in metric.terms:
            for c in self.components_by_location[term.taxonomy_category] if c.match(metric.filter) :
                    ...
```
Note that : 
- I renamed **components_by_taxonomy_category** by **components_by_location**, a bit shorter
- **components_by_location** was turned into a `dict[str, list[Component]`, instead of a `dict[str, dict[str, Component]`. Why ? Because the internal **dict** contains pairs (id, component), which is redundant : `component.id` is enough.

gems_views_builder/input/component.py
```diff
@@ -0,0 +1,152 @@
...
+        return str(self.raw_component.model.id).rsplit(".", 1)[-1]
+
+    @property
+    def properties(self) -> dict[str, str]:
+        return cast(dict[str, str], self.raw_component.properties)
+
+    def set_taxonomy_category(self, taxonomy_category: str) -> None:
+        self.taxonomy_category = taxonomy_category
+
+    def get_location(self, location_ports: str | tuple[str, ...] | None) -> str | tuple[str, ...]:
```
guilpier-code: About **get_location(...)** : 
- we may not need this method as a public one, given [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/78#discussion_r3538119334), but we might need it as a private one (so used internally to build data useful for the component to hold : see the comment linked above).
- It fetches, for each input port, the id of the unique component (= location) connected to the current component through the port. So the name is kind of misleading. So get_locations (with an **"s"**) seems a bit better.
- This method should return a **tuple**. It returns a collection of locations ("receiving" component ids). We should make this returned type as intuitive and simple regarding our business field. Same for its argument **location_ports** : its type is complicated (`str | tuple[str, ...] | None`). The tuple seems to be the more natural and intuitive type. **location_ports**'s current type directly comes from the yaml module loading operation. But we don't have to keep it. After loading, we should convert data into types that suits us (our business concerns)

**Caution** : this comment was changed a bit. So following comment by @dusanparipovic may seem inaccurate, but it used to be.

gems_views_builder/metrics_structure_builder.py
aoustry: @dusanparipovic @guilpier-code 

I don't understand your choice of the terminology `components_by_location` here @guilpier-code . I think there is a misunderstanding.

- This is not a matter of location here, this is a matter of **taxonomy_category** of the "contributors" to this metrics'term (here contributors are components).
- We should not use the terminology location at this line of the code, since it means something and this notion of location is used later in the process.

If you want a shorter terminology, we can plan a future PR where we'll replace "taxonomy-category" by "taxon", as you suggested.

dusanparipovic: # Dictionary Structure
Nested ID removal: Agreed to remove the nested ID from the dictionary structure. There is no need for O(1) access within the loop, so this simplification is acceptable.
Dictionary naming: Agreed with Antoine that the dictionary name could remain component_by_taxonomy_category. Alternatively, it could simply be named components, since access would look like 
```python 
components[term.taxonomy_category]
```
- making it self-evident that the intent is to retrieve components belonging to a specific taxonomy category.
@aoustry @guilpier-code

guilpier-code: ok to ban **location** from the name.
**components_by_taxon** seems clear enough (**taxon** does exist, and equals **taxonomy category**).
The short **components** misses something I think, but it's shorter, no doubt about it. Except if you already know,  when you see this dictionary you can wonder : what is components ? Is it the list of components of the system ? and you take a bit of extra time to answer. 
The rule we should respect is the **rule of least astonishment**. There should be no ambiguity, no question to ask.

gems_views_builder/input/component.py
dusanparipovic: I partly agree with you. In fact, we're looking for connected components, and the name is fine, but later in the process of building the metric structure table, we use it as location - because it represents the location where our component contributes. That's why `get_location` sounds more familiar to me. 
I agree that we make it **private**, and also that we could always return a **tuple** - there's no issue with that.
Regarding location_ports I agree, I'll see what I can do. 
@aoustry What do you think ?

guilpier-code: ok, I amended a bit my above comment. you're right : if this method turns out to be still useful,  I would just append an "s"  to the name:  **get_locations**

gems_views_builder/metrics_structure_builder.py
```diff
@@ -74,45 +49,27 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
-                                "output": term.output_id,
-                                "weight_output_id": 1,
-                            }
-                        )
-                    else:
-                        logging.debug(
-                            f"[{metric.id}] Component {component_id!r} did not match metric filter and was skipped"
-                        )
 
+            for c in self.components_by_taxono[term.taxonomy_category]:
```
guilpier-code: Typo : `self.components_by_taxono `==> `self.components_by_taxon`

gems_views_builder/metrics_structure_builder.py
```diff
@@ -12,59 +12,34 @@
...
+        components_by_taxono: dict[str, list[Component]],  # taxonomy category -> components
     ) -> None:
-        self.system = system
-        self.model_library = model_library
         self.location_taxonomy_category = location_taxonomy_category
+        self.components_by_taxono = components_by_taxono  # this is mainly for operating
 
-    def _location_component_matches_taxonomy_category(self, location_component_id: str) -> bool:
-        """Return True when the located component's model belongs to the view location taxonomy category."""
+    def _location_components_match_taxonomy_category(self, location_components: str | tuple[str, ...]) -> None:
```
guilpier-code: `def _location_components_match_taxonomy_category(...)`

According to [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/78#discussion_r3538119334), this function should disappear.

Indeed, just after loading the input data (especially **Library** + **System**), we should : 
1. check that, for a emitting component, and any of its ports, there is one unique receiving component (= a location)
2.  build the dictionary mentioned in the above comment for each emitting component

But to do this, each component must be able to know : 
- if it's an emitting component
- for any emitting component, and any of its ports, who is the (unique) component connected behind the port (which is the receiver, and also a location) ?

So how do we access these piece of information ?
There may be a better solution, but I see one : 
- Any component should know about its associated model (model are not used yet in **GVB**)
- if the associated model contains a non empty **port-field-definitions** part, it's an emitter (or contributor)

aoustry: @guilpier-code what about deleting this validation function in this PR, and opening an Issue for a future validation step to add in another part of the code ? Just to speed up the merge of this PR which is blocking for parallel works...

dusanparipovic: @aoustry I'm working on it

gems_views_builder/common.py
```diff
@@ -58,3 +67,28 @@ def configure_logging(verbose: bool = False, log_dir: Path | None = None) -> Non
     logger.addHandler(file_handler)
     logger.addHandler(console_handler)
     logger.setLevel(logging.DEBUG if verbose else logging.INFO)
+
+
+def preprocess_system_components(
```
guilpier-code: Here we roughly do 3 separate and quite different things : 
1. create a component list
2. enrich every component with new data
3. make a dictionary that groups components by taxon

So we could split **preprocess_system_components** in 3 : 
```python
components = create_components(input_data.system.components)
enrich_components(components, input_data.system, input_data.library)
components_by_taxon = group_components_by_taxon(components)
```
Note that : 
- the name **preprocess** does not tell much on what the function does. Indeed, it does multiple things, so it's hard to find a name for it. 
- The suggestion is to remove this function and replace it with 3 functions calls
-  the suggested split makes each new functions much easier to name
- regarding **enrich_components** : we can easily see that there are 2 parts 
  - give components a taxon
  - give components connections with other components
  
  So we could make **enrich_components** function call 2 other functions : **supply_components_with_taxons** and **supply_components_with_connexions**

tests/conftest.py
```diff
@@ -16,11 +16,39 @@
...
+)
+from gems_views_builder.loader import Loader
+from gems_views_builder.metrics_structure_builder import MetricStructureTableBuilder
+from gems_views_builder.view import ViewBuilder
 
 RESOURCES_TEST_FILES_ROOT = Path(__file__).resolve().parent.parent / "resources"
 TEST_INPUTS_PATH = RESOURCES_TEST_FILES_ROOT / "tests_inputs"
 
 
+def build_view_builder(dataset_dir: Path) -> ViewBuilder:
```
guilpier-code: This can be considered as code duplication : we could remove the call to : 
```py
StudyLayoutValidator(input_dir).validate()
```
from the top **run()** function, and we're good.
After all, this validation is not yet loading, just check every expected input file is available. 
So it's not really part of **running** something. 
Maybe **run** could be renamed into **run_view_building_process**, more clear and specific.

gems_views_builder/__main__.py
dusanparipovic: To be clear I want to call it `scope_taxon_category` it's much clear for my opinion, what do you think ?

gems_views_builder/common.py
```diff
@@ -58,3 +67,27 @@ def configure_logging(verbose: bool = False, log_dir: Path | None = None) -> Non
     logger.addHandler(file_handler)
     logger.addHandler(console_handler)
     logger.setLevel(logging.DEBUG if verbose else logging.INFO)
+
+
+def create_components(gemspy_components: list[GemsPyComponent]) -> list[Component]:
```
guilpier-code: All these functions (creation and supply functions) are about components.
They could be moved either to **components.py** or better, into a new **components_factory.py** (**components.py** would contain only class **Component**)

We could move also into **components_factory.py** : 
- group_components_by_taxon
- save_component_port_connections
- endpoint
- build_component_port_connections
- Particular case of **find_components_taxonomy_categories** : it does exactly what **supply_components_with_taxonomy_categories** is supposed to do. Move the body of the first to the second, and remove the first.

guilpier-code: Same with **compute_component_locations(...)** and **supply_components_with_locations(...)** : they are the same function. Remove the first, leave the second.

gems_views_builder/metrics_structure_builder.py
guilpier-code: At this stage (commit id : 7b73387e5b4f5e49eabc941bb6d1610258e92c97), it seems good !

But what if, while giving input components a location, a component is an area (that is a component receiving contribution from other connected component) ?

In this case, we may have a problem : these kind of components are generally connected to multiple emitting components. So, when giving input components a location associated to a couple (port, taxon), we may get multiple locations for these "receiving" components, not just a unique one, and so an unwanted exception would be raised.
If I'm right, we have to identify emitting components, and apply this check to them only.

Please also see [this connected comment / questionning ](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/78#discussion_r3559259742)

What do you think ?

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,193 @@
...
+    - more than one: a genuine inconsistency, raised immediately rather than later during table
+      building.
+    """
+    logging.info(f"Resolving component locations for taxonomy category {taxonomy_category!r}")
+    components_by_id = {component.id: component for component in components}
+    for component in components:
+        for port_id, peer_ids in component.connections.items():
+            matching_peers = [
+                peer_id for peer_id in peer_ids if components_by_id[peer_id].taxonomy_category == taxonomy_category
+            ]
```
guilpier-code: Please read [this comment](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/78#discussion_r3559259742) first, this one is a consequence.

If referenced comment is right, this is how function **compute_component_locations** could look like :  

```python
for c in components if c.is_contributor() :
    for connection in c.connections : 
        if len(connection.components()) > 1:
            # raising exception
        if len(connection.components()) == 1:
            connected = connection.components()[0]
            component.locations[(connection.port, connected.taxonomy_category)] = connected.id
```
Note that, to have this code, we would need a class **Connection**, containing a port and a list of associated components.

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,193 @@
...
+        if c1 == c2:
+            continue
+
+        component_port_connections[c1][p1].add(c2)
+        component_port_connections[c2][p2].add(c1)
+
+    logging.info(f"Built component-port connection index with {len(component_port_connections)} entry(ies)")
+    return component_port_connections
+
+
```
guilpier-code, line 95: About `def compute_component_locations(components, taxonomy_category)` : 

I have a doubt on what does this function exactly. 
So this comment is more a question than a affirmative comment.

My doubt comes from the second argument : **taxonomy_category**
This argument is actually the **config view scope location**.
Is it required here ?
Indeed, here we're supposed to give any component a location.
But to do that, we may only need the information contained in **System** and **Library**, I'm not sure we require the **view definition** (and its **location**) for this.

A component can be of 2 types : 
- it's a receiver : in that case, this component is its own location (areas, links)
- it's a emitter (or a contributor) : in this case, it's connected to one receiver components by port. It may be connected to multiple receiver components, but to different ports each time.

To me, the mission of this function is to give each component a location independently from the existence of a view.

**So where am I wrong ?**

tests/conftest.py
dusanparipovic: Done, see `__main__.py`

gems_views_builder/input/component/component_factory.py
dusanparipovic: > * It may be connected to multiple receiver components, but to different ports each time.

Actually we could have component which could be connected to several other components through the same port.
Example: area connected to several generators through port. 
But for GVB for a given port used as location for Metric, a component should have unique peer, if not than "location" doesn't make sense.

dusanparipovic: > To me, the mission of this function is to give each component a location independently from the existence of a view.

As you said earlier we don't want to break process of building simulation table later with raising error, so practically what we aim to do is be sure before building simulation table that we have already everything structured. (metric structure table process is then routine,just computation)
If you take a look at this: 
```python
matching_peers = [peer_id for peer_id in peer_ids if components_by_id[peer_id].taxonomy_category ==taxonomy_category]
```
So what we want to be sure that all locations are of mentioned taxonomy category in view config.
If we have multiple peers which are same taxonomy category, but if not we will regularly save it.
Later we will use this to get all locations which are in the `scope`

gems_views_builder/input/component/component_factory.py
dusanparipovic: For considerations:
Save all connections for each component and remove is_contributor check. 
Why? 
When we introduce multiple views we can reuse this connections over whole pipeline process for multiple views. 
What we currently do is that we check does our component/location component satisfy needs for scope, and now from my point of view that isn't something what we want to achieve, only what we want to be sure that we don't have multiple components of same taxonomy category as connection.
If this is case we can raise error here. 
Later we can perform check on connections do we match taxonomy-category from the scope, why this is safe, because we're sure that we won't have more than one peer. 
We will save multiple things which we don't need at this stage but soon when we introduce multiple views this will be very useful. 
@guilpier-code Please share your thoughts.

gems_views_builder/input/component/component.py
```diff
@@ -0,0 +1,89 @@
...
+@dataclass
+class Component:
+    """
+    Each component hold raw data from real component used for building additional fields
+    Taxonomy category
+    Connections to other components
+    """
+
+    raw_component: GemsPyComponent
+    taxonomy_category: str | None = None
```
aoustry, line 20: Is it the taxonomy_category of the GemsPy component itself? Or the taxonomy category of the ViewConfig (scope.location.taxonomy_category)?

gems_views_builder/input/component/component.py
```diff
@@ -0,0 +1,89 @@
...
+    def properties(self) -> dict[str, str]:
+        return cast(dict[str, str], self.raw_component.properties)
+
+    def set_taxonomy_category(self, taxonomy_category: str) -> None:
+        self.taxonomy_category = taxonomy_category
+
+    def is_located_at(self, location_ports: tuple[str, ...] | None, taxonomy_category: str) -> bool:
+        """Whether every port in ``location_ports`` has a resolved location for ``taxonomy_category``.
+
+        ``location_ports`` of ``None`` means the component is its own location: always true.
```
aoustry: Yes, but do you check that the component is allowed to be its own location, i.e. that its taxonomy_category matches the taxonomy_category of the View (scope.location.taxonomy_category) ?

gems_views_builder/input/component/component.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        if location_ports is None:
+            return True
```
aoustry: As explained about, I think you should check here that self.taxonomy_category == taxonomy_category

gems_views_builder/input/component/component.py
aoustry: (but maybe this is not the right place to do assertion checks ?)

gems_views_builder/input/component/component.py
dusanparipovic: Taxonomy category of component. 
If you take a look into `__main__.py` and function `supply_components_with_taxonomy_categories` you'll see that this function will fill this field for each component.

gems_views_builder/input/component/component.py
dusanparipovic: @aoustry 
We precompute component locations before building metric structure tables.

1. `supply_components_with_port_connections` stores component peer connections by port.
2. `supply_components_with_locations` scans those connections and stores only peers belonging to the view scope taxonomy category.
3. If a port has more than one peer in that category, we raise before computation starts.

Then, during metric structure building, `is_located_at(location_ports, scope_taxon_category)` checks whether all requested ports already have a resolved location. If yes, `formatted_locations(location_ports, scope_taxon_category)` uses the same parameters to read and format those precomputed locations as `metric_location`.

So at this stage we no longer resolve or validate connections deeply. We only check presence and read values that were already proven unique and part of the expected taxonomy category. This upfront processing avoids breaking in the middle or at the end of a large run.

gems_views_builder/input/component/component.py
dusanparipovic: [Answer](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/pull/78/changes#r3585476024)

gems_views_builder/input/component/component.py
guilpier-code: > Yes, but do you check that the component is allowed to be its own location, i.e. that its taxonomy_category matches the taxonomy_category of the View (scope.location.taxonomy_category) ?

As @dusanparipovic says, function **supply_components_with_locations** gives, for a given component, a unique location to each pair `(port, cfg_view_taxon)`.

But this tends to convince me that we need to know, in this function, if a component is **contributor** or **receiver** in order to give the component the right location : either itself if it's a receiver, or another one otherwise.

gems_views_builder/input/component/component_factory.py
guilpier-code: ok, it seems my understanding was inaccurate : I thought that, for a component, a location was independent from a view, which is not the case.

tests/conftest.py
guilpier-code: Actually, I was also encouraging you to avoid code duplication : 
`def build_view_builder` looks very much like **run_view_building_process(...)**, so I was wondering if we could remove **build_view_builder** from **conftest.py**, and use **run_view_building_process(...)** instead.
It seems possible.

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,124 @@
...
+def group_components_by_taxon(components: list[Component]) -> dict[str, list[Component]]:
+    """Group components by taxonomy category. Requires ``supply_components_with_taxonomy_categories`` first."""
+    components_by_taxon: dict[str, list[Component]] = defaultdict(list)
+    for component in components:
+        components_by_taxon[cast(str, component.taxonomy_category)].append(component)
+    return components_by_taxon
+
+
+def save_component_port_connections(
```
guilpier-code: **save_component_port_connections(...)** : we could rename this function into **supply_components_with_port_connections(...)**.
Indeed, using **save**, we could think that we save component connections on disk.
The suggested name : 
- is more accurate
- looks like other functions in the same file `supply_components` functions, so no surprise, we're being cohesive.

gems_views_builder/input/component/component_factory.py
guilpier-code: Actually I misunderstood that we need view's **taxonomy_category** to give locations to components.
So on that point you're right.

guilpier-code: On the need for contributor / receiver concept, it's still unclear but it's about to be decided.

gems_views_builder/input/component/component_factory.py
aoustry: @dusanparipovic

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,124 @@
...
+    logging.info(f"Built component-port connection index with {len(component_port_connections)} entry(ies)")
+    return component_port_connections
+
+
+def supply_components_with_port_connections(components: list[Component], system_connections: list[Any]) -> None:
+    component_port_connections = build_component_port_connections(system_connections)
+    save_component_port_connections(components, component_port_connections)
+
+
+def supply_components_with_locations(components: list[Component], taxonomy_category: str) -> None:
```
aoustry: This function is buggy. I suggest to recode it as follows

```def supply_components_with_locations(components_by_taxon: dict[str, list[Component]], ViewConfig, List[Catalogs]):

for m in ViewConfig.metrics:
     for t in m.terms:
          for c in components_by_taxon[t.taxonomy_category]:
                   if location_port == None
                                    if c.taxonomy_category !=t.taxonomy_category
                                              raise ValueError(  f"Component {c.id!r} cannot serve as metric location because it does not belong to taxonomy category {t.taxonomy_category!r}
                                    else:
                                          c.locations[(None,t.taxonomy_category)] = c.id

                   else :
                             connected_components_through_location_port = [ peer for peer in component_port_connections[c][t.location_port] if peer.taxonomy_category == t.taxonomy_category]
                             if len(connected_components_through_location_port ) != 1:
                                     raise ValueError(
                    f"Component {c.id!r} port {t.location_port} has {len(connected_components_through_location_port )} peers "
                    f"belonging to taxonomy category {t.taxonomy_category!r}, expected exactly one"
                )
                              else:
                                    c.locations[(t.location_port, t.taxonomy_group)] = connected_components_through_locating_port[0].id

Comment (aoustry) [reply gems_views_builder/input/component/component_factory.py]: @dusanparipovic : please note that in my pseudocode, `component_port_connections[component_id][port_id]` provide the list of components connected to `component_id` through port `port_id`. I let you replace this by the proper object that stores this information.

This was not very clear to me how c.connections could provide this information (filtering the peers by port).

File: gems_views_builder/metrics_structure_builder.py [hunk 21]
```diff
@@ -74,45 +36,17 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
-                                "weight_output_id": 1,
-                            }
-                        )
-                    else:
-                        logging.debug(
-                            f"[{metric.id}] Component {component_id!r} did not match metric filter and was skipped"
-                        )
 
+            for c in self.components_by_taxon[term.taxonomy_category]:
+                if c.match(metric.filter) and c.is_located_at(term.location_ports, self.scope_taxon_category):
```
aoustry: I think that the call to `c.is_located_at` is now pointless.

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,139 @@
...
+      - more than one peer of the same category → raise (ambiguous location).
+
+    Later read via ``Component.is_located_at`` / ``Component.resolve_location``, which both
+    look up ``(location_port, location_taxonomy_category)`` in ``Component.locations``.
+    """
+    for metric in metrics:
+        for term in metric.terms:
+            for c in components_by_taxon[term.taxonomy_category]:
+                if term.location_port is None:
+                    c.locations[(None, location_taxonomy_category)] = c.id
```
aoustry, line 128: A check that `term.taxonomy_category==location_taxonomy_category` is lacking here

gems_views_builder/input/component/connection.py
```diff
@@ -0,0 +1,18 @@
...
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from gems_views_builder.input.component.component import Component
+
+
+@dataclass
+class ConnectionThroughPort:
```
aoustry:
```suggestion
class ConnectionsThroughPort:
```

Please take care of details

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,139 @@
...
+                else:
+                    supply_component_with_location(c, term.location_port)
+
+
+def supply_component_with_location(component: Component, location_port: str) -> None:
+    peers_by_category: dict[str, list[Component]] = defaultdict(list)
+    for peer in component.connections.get_components(location_port):
+        peers_by_category[cast(str, peer.taxonomy_category)].append(peer)
```
aoustry: You can simplify: 

- you can get rid of peers_by_category
- you should have just one unique peer through the location_port (in total, regardless of the taxonomy category). If not, raise an error.
- Then you should check that the taxonomy category of this unique peer is indeed matching the location_taxonomy_category

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,139 @@
...
+    Later read via ``Component.is_located_at`` / ``Component.resolve_location``, which both
+    look up ``(location_port, location_taxonomy_category)`` in ``Component.locations``.
+    """
+    for metric in metrics:
+        for term in metric.terms:
+            for c in components_by_taxon[term.taxonomy_category]:
+                if term.location_port is None:
+                    c.locations[(None, location_taxonomy_category)] = c.id
+                else:
+                    supply_component_with_location(c, term.location_port)
```
aoustry:
```suggestion
                    supply_component_with_location(c, term.location_port, location_taxonomy_category)
```

gems_views_builder/input/component/component_factory.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def supply_component_with_location(component: Component, location_port: str) -> None:
```
aoustry:
```suggestion
def supply_component_with_location(component: Component, location_port: str, location_taxonomy_category: str) -> None:
```

gems_views_builder/metrics_structure_builder.py
```diff
@@ -74,45 +36,17 @@ def build(self, metric: Metric) -> MetricStructureTable:
...
-                                "weight_output_id": 1,
-                            }
-                        )
-                    else:
-                        logging.debug(
-                            f"[{metric.id}] Component {component_id!r} did not match metric filter and was skipped"
-                        )
 
+            for c in self.components_by_taxon[term.taxonomy_category]:
+                if c.match(metric.filter) and c.is_located_at(term.location_port, self.location_taxonomy_category):
```
aoustry, line 41: I'm repeating: is_located_at is now useless here.

AUTHORS.txt
```diff
@@ -1,5 +1,6 @@
-Antoine Oustry 
+Antoine Oustry
 Dusan Paripovic
+Guillame Pierre
```
aoustry:
```suggestion
Guillaume Pierre
```

PR #79, opened 2026-07-09
feat(view config): add taxonomy field and consistency check

### LINKED ISSUE #57: feat: add the taxonomy field in the view-config
- [x] Any view-config should have a unique field "taxonomy", that points towards a taxonomy name.
- [x] The scope.location-taxonomy-category should belong to this taxonomy.
- [x] When checking the consistency of a catalog with a view config, one should check that 1/ they point towards the same taxonomy ; 2/ the location taxonomy-category is the same in the catalog and the view config.

dusanparipovic:


gems_views_builder/__main__.py
```diff
@@ -37,6 +40,13 @@ def run(input_dir: Path, view_sinker: ViewSinker) -> None:
     # # If everything is ok, load pipeline input
     input_data = Loader(input_dir).load()
 
+    # # This could be grouped,now main is messy
```
aoustry: Yes indeed! We need here just a single call to a validation method, that executes all validations

gems_views_builder/__main__.py
```diff
@@ -37,6 +40,13 @@ def run(input_dir: Path, view_sinker: ViewSinker) -> None:
...
+    ViewConfigTaxonomyValidator(input_data.taxonomy, input_data.view_config).validate()
+    # # Validate catalogs against taxonomy and view config
+    catalogs = load_catalogs(input_dir, input_data.view_config.catalog_ids)
```
guilpier-code:
```py
catalogs = load_catalogs(input_dir, input_data.view_config.catalog_ids) : 
```
We load catalogs twice (from here, in main program, and also when executing 
`input_data = Loader(input_dir).load()` a bit earlier).

resources/tests_inputs/filtering_and_breakdown/view_config.yml
```diff
@@ -4,16 +4,20 @@ view:
...
+    - calendar:
+        id: calendar_file # calendar_file.csv
 
   aggregation:
     - time: hour
 
   catalog: #Pointer to "catalogs" = "Business Metrics definitions files".
       - id: catalog
 
+  taxonomy:
```
guilpier-code: If the **taxonomy** specified here is the one we want to use a category of in the
scope -> location -> taxonomy-category,
then it would be logical that the specified  taxonomy  comes first in **view_config.yml** (instead of after).

Is is not ?

gems_views_builder/input/view_config.py
guilpier-code, line 1: Question asked here : 
- because I can't see where else I could ask it
- to be answered some day, not necessarily here ans now (in some other ticket / PR)

Its about **catalogs** and **metrics** (inside a **catalog**) : 
These 2 collections are dictionaries in our **GVB**.
These dictionaries associated a catalog id to a catalog which already holds this id.   
So there is a kind of redundancy.
Could catalogs be a simple list of catalogs ?
Could a catalog (of class Catalog) hold a simple list of Metrics (instead of a dictionary currently) ?

If the answer is positive, it would make code a bit simpler and natural.

resources/tests_inputs/filtering_and_breakdown/view_config.yml
dusanparipovic: Very nice catch, I'll change pydantic modele for vc order and input test files.
But important remark, **order** for current pydantic model doesn't matter.

gems_views_builder/input/view_config.py
dusanparipovic: Agree for catalog list, changed. We could simplify to load list instead of dictionary this dictionary were used for old implementation approach of view builder, nice catch.

Why Catalog stores metrics as a dict, not a list
Catalog loads all metrics from file into memory. Later, we only need to pull the specific metrics referenced in a view config - not all of them.
If metrics were stored as a list, finding one metric by ID means scanning the list: O(N) per lookup, where N = total metrics in the catalog. Since a view config references M metrics, looking up all of them costs O(N × M).
If metrics are stored as a dict keyed by metric ID, each lookup is O(1), so pulling all M referenced metrics costs constant time.

gems_views_builder/__main__.py
dusanparipovic: Resolved,now loader will return `InputData` and `list[catalogs]` indeed we don't want to send catalogs with InputData because that is something what we don't expect to be passed to `ViewBuilder` object.(We don't need it)

gems_views_builder/input/view_config.py
```diff
@@ -119,7 +119,7 @@ def load_view_config(config_file_path: Path) -> ViewConfig:
     raw_view_config = load_raw_view_config_file(config_file_path)
     RawViewConfigValidator(raw_view_config).validate()
 
-    location_taxonomy_category = next(
+    scope_taxon_category = next(
```
aoustry: @dusanparipovic Please undo the change `location_taxonomy_category`-> `scope_taxon_category` (for all files impacted by commit 108c933). For two reasons:

- The notion of "location" is necessary here, since it relates to the location ports in the terms. 
- `taxon` means `taxonomy_category` in English, so `taxon_category` is a pleonasm. In the future, we may replace `taxonomy_category` by `taxon`, but for all occurences (note only for this one).

gems_views_builder/input/view_config.py
```diff
@@ -33,9 +33,17 @@ class TimeAggregation(Enum):
     YEAR = "year"
 
 
+class Location(ViewBuilderBasedModel):
```
aoustry: Maybe LocationCategory ? Because if the future we may have something like LocationFilter (for spatial filtering)

gems_views_builder/input/view_config.py
dusanparipovic: Done!

gems_views_builder/input/view_config.py
dusanparipovic: Yes right, in fact I called it LocationAggregation for spatial filtering

gems_views_builder/input/view_config.py
```diff
@@ -64,16 +77,23 @@ class ViewConfig:
...
-        for catalog_id in metric_ids_by_catalog:
-            for metric_id in metric_ids_by_catalog[catalog_id]:
-                self.metrics.append(catalogs[catalog_id].get_metric(metric_id))
+        for catalog in catalogs:
+            if catalog.id not in metric_ids_by_catalog:
+                raise ValueError(
+                    f"Catalog {catalog.id!r} has no metrics referenced in the view config. "
+                    f"Metric refs must use the catalog id from the catalog file "
+                    f"(expected prefixes: {sorted(metric_ids_by_catalog)})"
+                )
```
aoustry: I am not sure it is necessary to use this error. We don't really care if a catalog is in fact not used.

docs/input_validation_diagram.md
```diff
@@ -0,0 +1,67 @@
...
+| Location taxonomy category exists | `view_config.location_taxonomy_category` is a category defined in `taxonomy.yml` |
+
+### `catalog.yml` → `view_config.yml`
+[`CatalogsViewConfigValidator`](../gems_views_builder/validation/catalog_view_config_validator.py)
+
+| Check | Rule |
+|---|---|
+| Taxonomy id match | `catalog.taxonomy == view_config.taxonomy_id` |
+| Location taxonomy category match | `catalog.location_taxonomy_category == view_config.location_taxonomy_category` |
+
```
aoustry, line 63: I realised that an important check is missing: that all the metrics listed in the ViewConfig are indeed defined in the Catalogs. 

Please update this .md file (diagram + text) and the python file gems_views_builder/validation/catalog_view_config_validator.py accordingly.

gems_views_builder/__main__.py
```diff
@@ -27,16 +27,21 @@
...
 
-    # # Load pipeline input
-    input_data = Loader(input_dir).load()
+    # # Validate study layout
+    StudyLayoutValidator(input_dir).validate()
+
+    # # If everything is ok, load pipeline input
+    input_data, catalogs = Loader(input_dir).load()
+
+    InputConsistencyValidator(catalogs, input_data).validate()
```
aoustry, line 44: Still, this is a bit counter-intuitive that Catalogs do not belong to InputData. Can you open a Github Issue for a future fix?

gems_views_builder/validation/catalog_view_config_validator.py
```diff
@@ -0,0 +1,47 @@
...
+                f"but view config {self.view_config.id!r} references taxonomy "
+                f"{self.view_config.taxonomy_id!r}"
+            )
+        if catalog.location_taxonomy_category != self.view_config.location_taxonomy_category:
+            raise ValueError(
+                f"Catalog {catalog.id!r} location taxonomy category "
+                f"{catalog.location_taxonomy_category!r} does not match view config "
+                f"{self.view_config.id!r} location taxonomy category "
+                f"{self.view_config.location_taxonomy_category!r}"
+            )
```
aoustry, line 52: As asked above, we should add in this class the following validation checks: 
- each metric of the ViewConfig is indeed defined in a Catalog.
- all the metrics should have different ids (even if they do not belong to the same catalog)

docs/input_validation_diagram.md
```diff
@@ -0,0 +1,67 @@
...
+## Orchestration
```
aoustry, line 64: Moreover, we should have the following checks on the set of Catalogs: all the metrics ids should be different to avoid collision. Indeed, during the treatment, the catalogs names are removed from the identification-> collisions may occur if we don't forbid the coexistence of `catalogA.metric1` and `catalogB.metric1`.

gems_views_builder/validation/catalogs_taxonomy_validator.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+
+    @staticmethod
+    def _item_ids(items: list[TaxonomyItem]) -> set[str]:
+        return {item.id for item in items}
+
+    def _allowed_output_ids(self, category: TaxonomyCategory) -> set[str]:
+        # | logical or(union) so resulting set is unique.
+        return self._item_ids(category.variables) | self._item_ids(category.extra_outputs)
+
```
aoustry, line 45: - Not sure to understand this code structure...

gems_views_builder/input/view_config.py
dusanparipovic: Right removed, also note that group metrics by catalog is now "public" indeed this is not c++ and we don't have true public/private but I removed `_` because I need this function for validation.

gems_views_builder/validation/catalog_view_config_validator.py
dusanparipovic: Done!

gems_views_builder/__main__.py
dusanparipovic: Yes indeed that's tiny weird but why catalog isn't part of the input data because `input data` object  is aimed to store things only used in pipeline later(View Build and Metric Structure Table Build) 
That was initial plan when we constructed this object, now my questions is do we need to store library and system also?(everything is hidden in component class) 
[Issue](https://github.com/AntaresSimulatorTeam/GEMS-ViewsBuilder/issues/88)

docs/input_validation_diagram.md
dusanparipovic: Done!

docs/input_validation_diagram.md
dusanparipovic: Done!

gems_views_builder/validation/catalogs_taxonomy_validator.py
dusanparipovic: Logical or in python is | , so this is not boolean or. A or B. 
So basically it will return union of variables and extra outputs

dusanparipovic: I changed method and called .union function for 2 sets, its more readable sorry for this fancy stuff.

dusanparipovic:
I'm closing this PR because it out of date. 
New #115

